Issue #3, opened 2024-04-06
Port static packages


senceryazici:
Move packages listed below, from [auv-software-legacy](https://github.com/itu-auv/auv-software-legacy) to [auv-software](https://github.com/itu-auv/auv-software)

Port these packages:
- `auv_logger` 
- `auv_cam` 

Create a branch with appropriate branch name, create Pull Request targeting `main`. Attach this issue to the PR you've created. Ask for reviews when you're done, before merging.

YunusDiler:
some directory names in auv-software-legacy are for old name of jetson which is nvidia@jetson, the new name is jetson@ubuntu. they should be changed

senceryazici:
This is already fixed. `auv_logger` is obsolete, and `auv_cam` is already ported. so goodbye.

PR #15, opened 2024-04-30
move auv_logger and auv_cam

canerdev:
#3 
created the necessary folders and moved the auv_logger and auv_cam from auv-software-legacy to corresponding folders in the auv-software

update #1 -> fixed the directory structure. added the expressed directives in docker build file.


senceryazici:
Installing the cv_bridge package via apt on the jetson breaks up locally installed opencv for some reason. We need to tell the rosdep to ignore the requirement for cv_bridge, just like in previous docker build file [here](https://github.com/itu-auv/auv-software-legacy/blob/c8df5a23099c86170df7ffaa5ba9454c51590bf6/.docker/tegra/Dockerfile.ros.noetic-base#L89-L90)

senceryazici:
start with adding 
```
        --skip-keys=libopencv-dev \
        --skip-keys=cv_bridge \
```

directives for rosdep in docker build file. 


This can mess up the desktop build as well since cv_bridge is meant to be installed on a native desktop system. Lets see if CI fails.

senceryazici:
Also let's follow directory structure in #7. `auv_cam` will go under `auv_hardware`

senceryazici:
@canerdev Hello :D

canerdev:
@senceryazici su siralar yogunum biraz. ilgilenecegim.

docker/Dockerfile.auv-base
```diff
@@ -18,4 +18,6 @@ RUN apt-get update && \
   ros-noetic-xacro \
   ros-noetic-robot-localization && \
   rm -rf /var/lib/apt/lists/*
-
+  rosdep install \
```
senceryazici, line 21: This needs to be before rm -et /var/lib/apt/lists line


Also after each command do not forget '&& \' instead of '\'

Correct version should be

```
rosdep ... \
  --skip-keys ... \
  --skip-keys ... && \
rm -rf /var/...

```

senceryazici (changes requested):

docker/Dockerfile.auv-base
canerdev: done

senceryazici:
It seems here there is a better way to record topics with regular expressions. Let's remove the `auv_logger` changes from this PR, and convert it to single `auv_cam`

canerdev:
> It seems here there is a better way to record topics with regular expressions. Let's remove the `auv_logger` changes from this PR, and convert it to single `auv_cam`

done

PR #24, opened 2024-06-01
add logging system

senceryazici:
Previous usage was implementing a way to obtain topics from a yaml file. With this approach we can filter the topics with regular expressions. I've excluded topics end with `image_raw`. Can you test this with a system where camera is running and make sure it does not record the `image_raw` topics, but record rest of them @canerdev

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+github: [senceryazici, ituauvteam]
+buy_me_a_coffee: ituauvteam
+custom: ["https://auv.itu.edu.tr/sponsorship.html"]
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -17,7 +17,6 @@ jobs:
       image_type: base
 
   native-ci:
-    needs: build-and-push-base-docker
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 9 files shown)


canerdev:
> Previous usage was implementing a way to obtain topics from a yaml file. With this approach we can filter the topics with regular expressions. I've excluded topics end with `image_raw`. Can you test this with a system where camera is running and make sure it does not record the `image_raw` topics, but record rest of them @canerdev

will be able to work on it after 1st week of june

senceryazici:
@YunusDiler proceed with hardware test on this please. I've excluded all topics that contain `image_raw`

canerdev:
> @YunusDiler proceed with hardware test on this please. I've excluded all topics that contain `image_raw`

i think this is done last week and it works as expected. @senceryazici

senceryazici:
The current implementation does not exclude "image_rect_color" etc etc. So we still have to find out every topic name that has type of "sensor_msgs/Image"

ozanhakantunca:
These are the topics in a bag file @senceryazici 

/turquoise/cameras/cam_front/image_color                                                     1 msg     : sensor_msgs/Image 
/turquoise/cameras/cam_front/image_mono                                                      4 msgs    : sensor_msgs/Image 
/turquoise/cameras/cam_front/image_rect                                                      3 msgs    : sensor_msgs/Image 
/turquoise/cameras/cam_front/image_rect_color                                                1 msg     : sensor_msgs/Image

senceryazici:
Looks like `.*\/image_[^\/]*$` is good

PR #45, opened 2024-08-03
added auv_navigation package and added path follower node

batuhanozer:
added auv_navigation package and added path follower node


senceryazici:
Let's close this PR. @batuhanozer

PR #62, opened 2024-11-02
Update pressure_odometry_node.py

### LINKED ISSUE #70: adjust sensor offset for bar30
The depth data corresponds the Z height of the bar30, not the base link. Use proper static transform from pressure link to the base link, and also consider orientation of the robot. The output should be the base link position at all times.

optimistprime21:
Closes #70 

Adjusted the sensor offset for bar30. It now transforms depth data from the pressure sensor into the base link's frame.


senceryazici (commented):
Changes look good but we need to make sure the tf loopback wouldn't be a problem.

senceryazici (approved):
Approving in terms of coding, but still need to pass hardware test.

eminmeydanoglu:
Closing this PR. We are live at #137 now

PR #65, opened 2024-11-02
convert object map tf server to c++

ozanhakantunca:


senceryazici:
Not sure if this has to be tested on the hardware. Changes seem to be only affecting the TF, so software testing is also okay. But there is still some work to do, review-wise. So not ready for merge.

PR #71, opened 2024-11-10
move auv_localization to auv_navigation meta-package

### LINKED ISSUE #72: restructure workspace
I think below workspace structure is good

```
auv_navigation/
  - auv_navigation
  - auv_mapping
  - auv_localization
```
this requires the `auv_localization` to be moved from root directory to `auv_navigation`

When doing so, make sure dependencies are correctly adjusted.

`auv_navigation` is a meta-package and in it's [package.xml](https://github.com/itu-auv/auv-software/blob/main/auv_navigation/auv_navigation/package.xml), `auv_localization` and `auv_mapping` should be added as a dependency.

ozanhakantunca:
Closes #72 

Moved auv_localization package under auv_navigation and set dependencies.


auv_navigation/auv_navigation/package.xml
```diff
@@ -11,8 +11,11 @@
 
   <buildtool_depend>catkin</buildtool_depend>
   <build_depend>auv_mapping</build_depend>
-  <build_export_depend>auv_mapping</build_export_depend>
```
senceryazici, line 14: just use
```xml
  <buildtool_depend>catkin</buildtool_depend>
  <depend>auv_localization</depend>
  <depend>auv_mapping</depend>
```

senceryazici (approved):
Approving with small request. Good work

PR #76, opened 2024-11-18
add support for WrenchStamped

melihokur17:
Updated the controller_ros.h to publish WrenchStamped message type instead of Wrench to be able to subscribe in thruster_manager_ros.h and deal with transformation to the body frame and using it as is. It was very challenging for me since it is my first project assignement and pr.  

Looking forward to get feedback or suggestions for any corrections.


senceryazici (changes requested):
Let's see how it performs

senceryazici:
Also conflicts must be resolved.

senceryazici:
Resolved conflicts, removed unrelated commits. @melihokur17 Still, the comments need to be addressed though

PR #82, opened 2024-12-04
create auv_common_sim and taluy_sim

talhabw:
I made changes to follow the same structure in `taluy_descriptions`
- move the sensor and actuator plugins to `auv_common_sim_descriptions` package.
- move the base and dynamics to `taluy_sim_descriptions` package
- fix the find commands in xacros accordingly


talhabw:
Should I also move the content of `taluy_sim_description` to `auv_sim_description` or is this structure better? I thought this makes more sense, since everything in the `taluy_sim_description` package is spesific to taluy.

senceryazici (approved):
Good job, there was only a small change, so I pushed that in order to save some time.

PR #86, opened 2024-12-06
add dvl velocity implementation to auv_sim

### LINKED ISSUE #84: add dvl plugin to simulation

talhabw:
Closes #84

I tried the DVL plugin from the uuv_simulator first, but the velocity values I got weren't consistent with ground truth. I think the the plugin wasn't handling the transformation between the base_link and the odom correctly, or not correctly considering the rotation of the DVL link. 

I tried giving a wrench in the Y axis and plotting Y axis speed we get from different nodes. Robot was approximately drawing a circle around a point, here is a screen recording* of these plots.

https://github.com/user-attachments/assets/babe0432-1495-48a5-9714-1fbad782e5d5

Since these velocity values are already published by gazebo, and not calculated by an acoustic simulation, I thought we could just use the values in `odometry_gt` (our ground truth topic). I added a 135 degree rotation to our simulation mock node to compensate for the rotation in [`dvl_odometry_node`](https://github.com/itu-auv/auv-software/blob/talha/sim-dvl-plugin/auv_navigation/auv_localization/scripts/dvl_odometry_node.py#L95). This implementation seems to work a lot better, since the DVL values are _really close_ to gt in real life conditions too. 

Here is a screen recording* of the approximately same movement happening with this implementation.

https://github.com/user-attachments/assets/610e4412-787a-45f6-84e0-b36e71480792

It would be great to see what you think about this.

*: I couldn't find a way to increase the time step high enough in plotjuggler to make this recordings just a screenshot, which would have been obviously better... they are also sped up 8x so it would be easier to examine


auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
```diff
@@ -40,12 +42,49 @@ class SimulationMockROS {
...
+    Eigen::Vector3d rotated_velocity = rotation_matrix * velocity_vector;
+
+    geometry_msgs::Twist velocity_msg;
+    velocity_msg.linear.x = rotated_velocity.x();
+    velocity_msg.linear.y = rotated_velocity.y();
+    velocity_msg.linear.z = rotated_velocity.z();
+    return velocity_msg;
+  }
+
+  geometry_msgs::Twist addNoiseToTwist(const geometry_msgs::Twist &input) {
```
senceryazici, line 71: Instead of making a copy, you could write to the given argument in place by removing const and leaving the ref (&). But I don't even think it will matter that much

auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
```diff
@@ -81,24 +120,32 @@ class SimulationMockROS {
...
     altitude_pub_.publish(altitude_msg);
 
+    latest_altitude_ = msg.range;
+  }
+
+  void dvlCallback(const nav_msgs::Odometry &msg) {
+    geometry_msgs::Twist velocity_raw_msg;
+    std_msgs::Bool is_valid_msg;
+
+    if (latest_altitude_ > 0.3 && dvl_enabled_) {
```
senceryazici, line 136: Lets make 0.3 either a rosparam or a constant st the top of the file. A constant would be sufficient.

auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
```diff
@@ -81,24 +120,32 @@ class SimulationMockROS {
...
-    velocity_stamped_msg.header.frame_id = "dvl_link";
-    velocity_stamped_msg.twist.twist = velocity_msg;
-    velocity_stamped_msg.twist.covariance = covariance;
-    velocity_stamped_pub_.publish(velocity_stamped_msg);
+    velocity_raw_pub_.publish(velocity_raw_msg);
+
+    // Dvl validity
+    is_valid_pub_.publish(is_valid_msg);
+
+    dvl_rate_.sleep();
```
senceryazici, line 148: Sleeping here would create queue problems, instead you should store the latest reading, and run your modifications on a different loop which rolls at dvl rate. Use ROS timer perhaps to detach a different process which dies not block the loop

auv_sim/auv_sim_description/taluy_sim_description/urdf/sensors.urdf.xacro
```diff
@@ -13,4 +13,6 @@
     topic_name="cameras/cam_bottom" />
   <xacro:xsens_mti_g710_sim name="${namespace}/base_link/imu" namespace="${namespace}"
     static_reference_frame="${inertial_reference_frame}" />
-</robot>
+  <xacro:altimeter_sim name="${namespace}/base_link/altimeter" parent="${namespace}/base_link"
```
senceryazici, line 16: Instead of attaching this to a different frame, you can use the dvl frame. Our dvl frame should be  oriented Z down if I'm Not mistaken. If that replicates your usage here maybe use that. If not, just set your parent to the dvl link and orient however you want and set it "altimeter" frame like you do here. No need to transform all the way from the origin. I'm not sure how our altitude message is sent. If it is sent with a header, then its frame id also should be matched, forcing you to mount this sensor exactly on whatever frame that is. Feel free to investigate this

auv_sim/auv_sim_description/taluy_sim_description/urdf/sensors.urdf.xacro
```diff
@@ -13,4 +13,6 @@
...
+    position="0.2 0 -0.11" orientation="0 0 0" />
```
senceryazici, line 17: In that case (described above) position will be 0 0 0, and only orientation will be provided

auv_sim/auv_sim_description/taluy_sim_description/urdf/taluy_sim.urdf.xacro
```diff
@@ -4,7 +4,7 @@
   <xacro:arg name="debug" default="0" />
   <xacro:arg name="namespace" default="taluy" />
   <xacro:arg name="control_rate" default="20" />
-  <xacro:arg name="inertial_reference_frame" default="world" />
+  <xacro:arg name="inertial_reference_frame" default="odom" />
```
senceryazici, line 7: This is the default argument and let's leave this as is. The spawn launch should be injecting the actual values that we use. So lets modify that. Even in that case should we be changing this to odom ?

senceryazici:
Oh btw
> I couldn't find a way to increase the time step high enough in plotjuggler to make this recordings just a screenshot, which would have been obviously better... they are also sped up 8x so it would be easier to examine

There should be a text box at the top left says 30. Try increasing that. That is what you referred to right? Time window?

senceryazici:
Also I just did a code review, I haven't been able review those videos, yet! I'll take a look at them and come back.

auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
talhabw: used reference instead of making a copy as you suggested

auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
talhabw: I understand why this would create a problem. I tried using a ROS timer to create a different callback with dvl rate, but it seems to have a problem. This is a log of the implementation in the current code, I will investigate this further.

```
[1733629144.805371388, 2.100000000]: TIMER CALLBACK
[1733629145.272659647, 2.250000000]: TIMER CALLBACK
[1733629145.428972508, 2.300000000]: TIMER CALLBACK
[1733629145.741480176, 2.400000000]: TIMER CALLBACK
[1733629146.052502991, 2.500000000]: TIMER CALLBACK
[1733629146.361058481, 2.600000000]: TIMER CALLBACK
[1733629146.673469617, 2.700000000]: TIMER CALLBACK
[1733629147.133962142, 2.850000000]: TIMER CALLBACK
[1733629147.445909725, 2.950000000]: TIMER CALLBACK
[1733629147.604225889, 3.000000000]: TIMER CALLBACK
```

auv_sim/auv_sim_description/taluy_sim_description/urdf/sensors.urdf.xacro
talhabw: This makes sense, now we use dvl frame. If my observations are right, the dvl frame is not oriented Z down, so I added the necessary orientation for the altimeter.

Altitude data is sent as float32, there is no header implementation. There is some hard coded offset in the node its used, maybe making an actual transformation using the dvl frame can be a future improvement in that node.

auv_sim/auv_sim_description/taluy_sim_description/urdf/taluy_sim.urdf.xacro
talhabw: Oversight on my part, didn't realize it was a default. I am not actually sure about if we should use world or odom, now it seem to didn't have any difference in my testing but we can investigate this further.

reverted this.

auv_sim/auv_sim_description/taluy_sim_description/urdf/sensors.urdf.xacro
senceryazici: Right good catch. It seems publishing the altitude from the dvl frame is a bit useless since the altitude information (if we wanted to use it) would be better with base to bottom distance. So I  don't know why I used the dvl altitude without the header frame but whatever, yours look good and reflect the current codebase. If we decide to improve the codebase in the future, we update this node too.

auv_sim/auv_sim_description/taluy_sim_description/urdf/taluy_sim.urdf.xacro
senceryazici: Right, even though the header frame is "world" all the twist information is still in the body frame. That frame only would effect the position, and for the time being we don't have anıt other pose reference frame. The odom will be published by the ekf node, so no problems as far as the dvl part is concerned.

auv_sim/auv_sim_bridge/include/auv_sim_bridge/simulation_mock_ros.h
senceryazici: Hmm, yeah that seems to be a bit odd. Maybe it is related to the RTF if your node is not using simulation time. Or even taking it back, maybe the timestamps that you see is the real time instead of the sim time, thus there is a difference. Anyways I dint want to point you into the wrong direction so feel free to debug this any way you like

talhabw:
I spent some time debugging the pose3d plugin we use to publish ground truth. I don't think its actually publishing the value we expect. There was also some misunderstanding on my part about how the transformation between dvl, base_link and odom frames are handled.

Latest code uses the dvl plugin from uuv_simulator and handles noise and rotation in the bridge node. We are also using the altitude from the same plugin so everything is simpler.

The first post and the videos are not really relevant anymore. Squash merge might be the best choice for this pr since I circled around the same implementation a few times.

senceryazici (approved):

PR #90, opened 2024-12-07
Better launch files

eminmeydanoglu:
improvements to simulation-related launch files to enhance clarity and convenience.

### DIFF
--- a/auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
+++ b/auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
@@ -2,10 +2,9 @@
 <launch>
   <arg name="gui" default="true" />
   <arg name="debug" default="false" />
-  <arg name="world_name"/>
+  <arg name="world_name" default="$(find auv_sim_description)/worlds/seabed.world" />
   <arg name="namespace" default="taluy" />
-  <arg name="velocity_control" default="true" />
-  <arg name="joy_id" default="0" />
+  <arg name="seabed_world" default="$(find auv_sim_description)/worlds/seabed.world" />
 
   <include file="$(find gazebo_ros)/launch/empty_world.launch">
     <arg name="world_name" value="$(arg world_name)" />
@@ -16,9 +15,9 @@
     <arg name="verbose" value="true" />
   </include>
 
-  <node if="$(eval arg('world_name') == '$(find auv_sim_description)/worlds/seabed.world')" 
-        name="publish_world_models" pkg="uuv_assistants" type="publish_world_models.py"
-        output="screen">
-    <rosparam command="load" file="$(find auv_sim_bringup)/config/world_models.yaml" />
-  </node>
+  <group if="$(eval arg('world_name') == arg('seabed_world'))">
+    <node name="publish_world_models" pkg="uuv_assistants" type="publish_world_models.py" output="screen">
+      <rosparam command="load" file="$(find auv_sim_bringup)/config/world_models.yaml" />
+    </node>
+  </group>
 </launch>
--- a/auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
+++ b/auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
@@ -2,7 +2,7 @@
 <launch>
   <arg name="gui" default="true" />
   <arg name="debug" default="false" />
-  <arg name="world_name" default="$(find auv_sim_description)/worlds/seabed.world" />
+  <arg name="world_name"/>
   <arg name="namespace" default="taluy" />
   <arg name="velocity_control" default="true" />
   <arg name="joy_id" default="0" />
@@ -16,8 +16,9 @@
     <arg name="verbose" value="true" />
   </include>
 
-  <node name="publish_world_models" pkg="uuv_assistants" type="publish_world_models.py"
-    output="screen">
+  <node if="$(eval arg('world_name') == '$(find auv_sim_description)/worlds/seabed.world')" 
+        name="publish_world_models" pkg="uuv_assistants" type="publish_world_models.py"
+        output="screen">
     <rosparam command="load" file="$(find auv_sim_bringup)/config/world_models.yaml" />
   </node>
 </launch>
--- a/auv_sim/auv_sim_bringup/launch/start_gazebo.launch
+++ b/auv_sim/auv_sim_bringup/launch/start_gazebo.launch
@@ -2,16 +2,14 @@
 <launch>
   <env name="GAZEBO_MODEL_PATH" value="$(find auv_sim_description)/models:${GAZEBO_MODEL_PATH}"/>
   <arg name="namespace" default="taluy" />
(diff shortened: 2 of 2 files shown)


auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
```diff
@@ -2,10 +2,9 @@
 <launch>
   <arg name="gui" default="true" />
   <arg name="debug" default="false" />
-  <arg name="world_name" default="$(find auv_sim_description)/worlds/seabed.world" />
+  <arg name="world_name" default="$(find auv_sim_description)/worlds/pool.world" />
```
eminmeydanoglu, line 5: Under normal circumstances, you want the pool world to be the default.

auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
```diff
@@ -16,8 +15,9 @@
...
-    <rosparam command="load" file="$(find auv_sim_bringup)/config/world_models.yaml" />
-  </node>
-</launch>
\ No newline at end of file
+  <group if="$(eval arg('world_name') == arg('seabed_world'))">
+    <node name="publish_world_models" pkg="uuv_assistants" type="publish_world_models.py" output="screen">
+      <rosparam command="load" file="$(find auv_sim_bringup)/config/world_models.yaml" />
+    </node>
+  </group>
+</launch>
```
eminmeydanoglu, line 23: The publish_world_models node is only relevant for the seabed world and is unnecessary for the pool environment. With this conditional check, the node will only execute when world_name is set to seabed_world

auv_sim/auv_sim_bringup/launch/start_gazebo.launch
```diff
@@ -1,16 +1,15 @@
 <?xml version="1.0"?>
 <launch>
+  <env name="GAZEBO_MODEL_PATH" value="$(find auv_sim_description)/models:${GAZEBO_MODEL_PATH}"/>
```
eminmeydanoglu, line 3: Set the gazebo model path correctly and automatically for convenience. Works as intended 👍

auv_sim/auv_sim_bringup/launch/start_gazebo.launch
```diff
@@ -1,16 +1,15 @@
...
+  <arg name="world_name" default="$(find auv_sim_description)/worlds/pool.world" />
   <arg name="control_rate" default="20.0" />
-  <arg name="use_ned_frame" default="false" />
   <arg name="reference_frame" value="world" />
-
   <arg name="use_gui" default="true" />
   <arg name="debug" default="false" />
-  <arg name="world_name" default="$(find auv_sim_description)/worlds/seabed.world" />
-  <arg name="velocity_control" default="true" />
-  <arg name="joy_id" default="0" />
```
eminmeydanoglu, line 12: Never use these two. Unnecessary.

auv_sim/auv_sim_bringup/launch/start_gazebo.launch
```diff
@@ -19,12 +18,10 @@
...
-    <arg name="gui" default="$(arg use_gui)" />
-    <arg name="debug" default="$(arg debug)" />
-    <arg name="world_name" default="$(arg world_name)" />
-    <arg name="namespace" default="$(arg namespace)" />
-    <arg name="velocity_control" default="$(arg velocity_control)" />
-    <arg name="joy_id" default="$(arg joy_id)" />
+    <arg name="gui" value="$(arg use_gui)" />
+    <arg name="debug" value="$(arg debug)" />
+    <arg name="world_name" value="$(arg world_name)" />
+    <arg name="namespace" value="$(arg namespace)" />
```
eminmeydanoglu, line 24: Why set default something that is already set default up top?
Switched default to value. Keeps things consistent and avoids duplication.

auv_sim/auv_sim_bringup/launch/start_gazebo.launch
senceryazici: Is this required? The package.xml already has export section and should be handling this automatically. Can you clean your build, remove this, and rely on the package.xml export and see if problem persists?

auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
senceryazici: I understand what you want to achieve here. But what if we have another world to add? Will we make this launch file have bunch of if else sections? Maybe we can add 

launch/inc/worlds/spawn_pool.launch
...

launch/inc/worlds/spawn_<world_name>.launch


Then we can really do, 

roslaunch auv_sim_bringup start.launch world:=seabed

Which will 

<include file="..... launch/inc/worlds/spawn_$(arg world).launch">

And in that folder we add all spawn_... for each world we want to use. 

I also think that this node you're referring can be included inside the world file just like the pool world, but I think editing world files and not being able to benefit from launch files' arg,find,if,else functions will make things harder eventually, thus keeping world specific launch file could be a good idea. 
 
Of course this could be investigated, no pressure. Maybe you just shake the tree and see what comes down?

auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
senceryazici: Which pool? :)

The third world I've mentioned below could be answer to this question. We will have 2 worlds, 1 for RoboSub 1 for the other competition?

senceryazici (commented):
Just left sone questions or some topics that can be discussed. There is not much of a review there, so not enforcing any idea

auv_sim/auv_sim_bringup/launch/start_gazebo.launch
eminmeydanoglu: You are referring to;

```xml
    <gazebo_ros gazebo_model_path="${prefix}/meshes" />
    <gazebo_ros gazebo_model_path="${prefix}/urdf" />
    <gazebo_ros plugin_path="${preffix}/lib" />
    <gazebo_ros gazebo_media_path="${preffix}" />
```
I tested these, and also tried different methods utilizing package.xml, but they don't seem to work. Best way I could find was setting the environment variable at .launch file. 

Here is a quote from ChatGPT explaining why package.xml probably won't work:

> Unfortunately, using just the package.xml file to set environment variables at runtime is not a standard approach in ROS 1. In ROS 1 (with catkin), environment variables are typically set through environment hooks, rather than directly through package.xml.
> Why package.xml doesn’t work directly for environment variables:
> In ROS 1, package.xml describes package metadata (name, version, dependencies, etc.) but does not provide a mechanism to dynamically modify environment variables at runtime after the user sources their workspace.

auv_sim/auv_sim_bringup/launch/inc/start_world.launch.xml
eminmeydanoglu: I've made some updates about this. You can check the other PR @melihokur17 started.

PR #91, opened 2024-12-09
Refactor and update gate trajectory logic and configurations

optimistprime21:
- Updated "start.launch" in "auv_control" in order to prevent "gate_trajectory_publisher" to run when starting the simulation.
- Adjusted "start.launch" in "auv_mapping" to align with new logic.
- Moved "gate_trajectory_publisher.py" and "prop_transform_publisher.py" from "auv_control" to "auv_mapping" to better organize scripts related to navigation and mapping.
- Modified "gate.py" in "auv_smach" to incorporate new trajectory handling.

After running the simulation, if you keep getting an error related to "gate_blue_arrow_link", this issue will be resolved once these changes are applied.


auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
```diff
@@ -29,7 +29,8 @@ def __init__(self):
         self.world_frame = "odom"
 
         self.selected_frame = "gate_blue_arrow_link"
-
+        self.set_enable_service = rospy.Service('taluy/set_gate_trajectory_enable', SetBool, self.handle_enable_service)
```
senceryazici, line 32: remove taluy here, this node will be placed in `<group ns="$(arg namespace)">` already

Also, rename the service to something like `set_transform_gate_trajectory`?

auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
```diff
@@ -148,10 +149,17 @@ def send_transform(self, transform):
                 f"Failed to set transform for {transform.child_frame_id}: {resp.message}"
             )
 
+    def handle_enable_service(self, req):
+        self.enable = req.data
+        rospy.loginfo(f"Set enable to {self.enable}")
```
senceryazici, line 154: f"Gate trajectory transform publish is set to: {self.enable}"

auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
```diff
@@ -148,10 +149,17 @@ def send_transform(self, transform):
...
+        return SetBoolResponse(success=True, message="Enable set to " + str(self.enable))
```
senceryazici, line 155: similar to above, `enable set` is not descriptive enough.

auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
senceryazici: also at the beggining there is a line:

```py

 self.set_object_transform_service = rospy.ServiceProxy(
            "/taluy/map/set_object_transform", SetObjectTransform
        )
```

this must be `map/set_object_transform` due to the similar reasons.

senceryazici: it would be even better if these can be obtained from ros params as well


```py
 self.frame1 = rospy.get_param("~gate_frame1", "gate_blue_arrow_link")
 self.frame2 = rospy.get_param("~gate_frame2", "gate_red_arrow_link")
 self.new_frame_away = rospy.get_param("~gate_exit_frame", "gate_exit")
 self.new_frame_closer = rospy.get_param("~gate_enterance_frame", "gate_enterance")
 self.selected_frame = rospy.get_param("~gate_selected_frame", "gate_blue_arrow_link")
 self.world_frame = rospy.get_param("~world_frame", "odom")
```

of course, launch files needs to be adjusted as well

auv_smach/src/auv_smach/gate.py
```diff
@@ -41,8 +50,21 @@ def __init__(self, gate_depth: float):
...
         )
 
+        # Service to control the enable state of the TransformServiceNode 
+        self.set_enable_service = rospy.ServiceProxy('taluy/set_gate_trajectory_enable', SetBool)
+
         # Open the container for adding states
         with self.state_machine:
+            
+            smach.StateMachine.add(
+                "ENABLE_GATE_TRAJECTORY",
```
senceryazici, line 60: ENABLE_GATE_TRAJECTORY_PUBLISHER

auv_smach/src/auv_smach/gate.py
```diff
@@ -31,6 +31,15 @@
 
 from auv_smach.initialize import DelayState
 
+class TransformServiceEnableState(smach_ros.ServiceState):
+    def __init__(self, req:bool):
+        smach_ros.ServiceState.__init__(
+            self,
+            "taluy/set_gate_trajectory_enable",
```
senceryazici, line 38: again, no taluy.

auv_smach/src/auv_smach/gate.py
```diff
@@ -41,8 +50,21 @@ def __init__(self, gate_depth: float):
             outcomes=["succeeded", "preempted", "aborted"]
         )
 
+        # Service to control the enable state of the TransformServiceNode 
+        self.set_enable_service = rospy.ServiceProxy('taluy/set_gate_trajectory_enable', SetBool)
```
senceryazici, line 54: I don't think this is required, so remove please.

auv_smach/src/auv_smach/gate.py
```diff
@@ -88,12 +110,23 @@ def __init__(self, gate_depth: float):
                 NavigateToFrameState(
                     "gate_enterance", "gate_exit", "gate_target", n_turns=-1
                 ),
+                transitions={
+                    "succeeded": "DISABLE_GATE_TRAJECTORY",
```
senceryazici, line 114: DISABLE_GATE_TRAJECTORY_PUBLISHER

auv_control/auv_control/config/default.yaml
```diff
@@ -23,7 +23,7 @@ model:
     - [0, 0, 0, 0, 0.0, 0]
     - [0, 0, 0, 0, 0, 7.98053]
 
-#kp: [0.0, 0.0, 1.5, 0.0, 0.0, 0.0, 1.0, 1.0, 1.0, 5.0, 0.0, 1.0]
```
senceryazici, line 26: this is irrelevant, so please remove

senceryazici (changes requested):

optimistprime21:
Made the requested changes. If there's anything you would like me to change again, I can do it.

senceryazici (dismissed):
Looks good to me, it would be good to check the logic simply by transforming 2 links from odom, and start the gate trajectory publisher and see if the entertainer and exit are created at selected frame. The state machine part looks good too.

senceryazici:
I think this needs some sort of a rebase :D

optimistprime21:
I might have done some mistakes during rebasing, I will resolve them as soon as possible.

talhabw:
@senceryazici your review told to remove `/taluy` namespace from the service state in `gate.py`, but this doesn't seem to be correct. We don't run the smach in the `/taluy` namespace (maybe we should, which could be a subject of another pr), so we need the use the namespace while defining service proxies. There are another examples of this in `common.py` file.

Requesting a review again just to be sure. This worked as it should on the simulation test after namespace fix.

auv_navigation/auv_mapping/launch/start.launch
```diff
@@ -7,10 +7,21 @@
...
+      <param name="gate_frame1" value="gate_blue_arrow_link" />
+      <param name="gate_frame2" value="gate_red_arrow_link" />
+      <param name="gate_exit_frame" value="gate_exit" />
+      <param name="gate_enterance_frame" value="gate_enterance" />
+      <param name="gate_selected_frame" value="gate_blue_arrow_link" />
+      <param name="world_frame" value="odom" />
+    </node>
+  </group>
+  <!--
+  <node pkg="auv_control" type="bin_transform_publisher.py" name="bin_transform_publisher" />
```
senceryazici, line 25: This can be removed right?

senceryazici (approved):
This needs to be tested on sim though. Or on the robot

optimistprime21:
Screenshot from 2025-01-04 15-48-28

I tested the code in the simulation. I published a TF from odom to gate_blue_arrow_link and gate_red_arrow_link, and then I ran the gate_trajectory_publisher. In the screenshot you can see the data related to the created gate_exit and gate_entrance frames in the bottom two terminals.

auv_navigation/auv_mapping/launch/start.launch
optimistprime21: Yes, I think so.

senceryazici:
Is this ready? if so rebase this please and request `re-review`. @optimistprime21 @talhabw

eminmeydanoglu:
Can we also solve #119 in this PR? @talhabw

senceryazici:
> Is this ready? if so rebase this please and request re-review. @optimistprime21 @talhabw

Up. What is the status here?

talhabw:
this passed hardware and simulation tests. @senceryazici waiting for your review

auv_smach/src/auv_smach/gate.py
```diff
@@ -32,6 +32,16 @@
 from auv_smach.initialize import DelayState
 
 
+class TransformServiceEnableState(smach_ros.ServiceState):
+    def __init__(self, req: bool):
+        smach_ros.ServiceState.__init__(
+            self,
+            "/taluy/set_transform_gate_trajectory",
```
senceryazici, line 39: no absolute topic naming

senceryazici (approved):
LGTM except one comment

eminmeydanoglu:
I think there are some bugs in this PR;
1. It is that the transforms won't disappear at "DISABLE_GATE_TRAJECTORY_PUBLISHER" state. And I guess this is expected, since there is no mechanism to remove them when enable becomes false! You might think they will disappear if we are not creating new ones, but notice we are using the set_object_transform_service, and I guess that is why simply not publishing the transforms aren't working. 
2. We can consider sending a empty transform. Sending an empty transform with the same child_frame_id should signal to the set_object_transform_service to remove the frames.
3. The order of smach states in gate.py looks flawed. The order is, "ENABLE_GATE_TRAJECTORY_PUBLISHER", "SET_GATE_DEPTH", "SET_ALIGN_CONTROLLER_TARGET", "DISABLE_GATE_TRAJECTORY_PUBLISHER", "NAVIGATE_TO_GATE_START", "NAVIGATE_TO_GATE_EXIT". Well but we need the gate entrance/exit frames specifically for the last two navigate_to_x states! If disabling would work, I think these states would not work, since they look up for these frames. gate_start state looks for gate_entrance, and gate_exit state looks for both. 
@talhabw @senceryazici

PR #95, opened 2024-12-14
Update joy function

ozanhakantunca:
The previous _joy_manager.py_ file was configured for an Xbox controller, which caused it to perform incorrect functions. It has been updated to work with the joystick used during testing.


auv_teleop/scripts/joy_manager.py
```diff
@@ -85,9 +85,9 @@ def joy_callback(self, msg):
         with self.lock:
             self.joy_data = msg
 
-            self.torpedo1_button_event.update(self.joy_data.buttons[0])
-            self.torpedo2_button_event.update(self.joy_data.buttons[1])
-            self.droper_button_event.update(self.joy_data.buttons[3])
+            self.torpedo1_button_event.update(self.joy_data.buttons[4])
```
senceryazici, line 88: Looks good, but I'm going to be annoying for the next 5 minutes :D 

why not make these ros param? for example


```yaml
buttons:
  launch_torpedo1: 5
  launch_torpedo2: 3
  drop_ball: 2


axes:
  surge:  # (forward)
    index: 1 # button index
    gain: 0.1 # scalar
  sway: # lateral
    index: 2
    gain : 0.5
  yaw:
    index: 3 
    gain: 0.7


```

senceryazici (changes requested):

senceryazici:
then you can load different config files for different joysticks

```xml

  ...
  <rosparam command="load" file="$(find auv_teleop)/config/psp4.yaml"/>
  ...
```

ozanhakantunca:
Yes, that makes sense; it’s a much better solution. I’ll be working on it. :)

senceryazici:
Btw hardware test of this PR could be as simple as observing the cmd_vel output with the joystick. There is no need to test this with vehicle in pool.

auv_teleop/config/psp4.yaml
```diff
@@ -0,0 +1,19 @@
+buttons:
```
senceryazici, line 1: Isn't it PS4? :D

ozanhakantunca:
Should I remove the README.md file in the package. Is it neccessary ? @senceryazici

senceryazici (approved):
Good work, this can be merged if tested.

PR #98, opened 2024-12-16
fix orientation control

talhabw:
Quaternion to Euler angle conversion function was not working properly. There was an issue about wrapping/normalizing angles, and it was returning incorrect values, which caused problems while calculating the errors for pid.

<details><summary></code>Here is a log of what we calculate as orientation error while the robot is stable and `desired orientation = current orientation</code></summary>
<p><pre>[ WARN] [1734376964.945058667, 79.440000000]: ERRORS: -3.13868  3.13726 -3.14159
[ WARN] [1734376965.017779235, 79.490000000]: ERRORS: -3.13943  3.13808 -3.14098
[ WARN] [1734376965.098415150, 79.540000000]: ERRORS: -0.00404835  0.00684969 -0.00184575
[ WARN] [1734376965.175873698, 79.591000000]: ERRORS: -0.00404835  0.00684969  1.9409e-06
[ WARN] [1734376965.245359836, 79.640000000]: ERRORS: -0.00255844 2.78368e-05   0.0032894
[ WARN] [1734376965.334101698, 79.691000000]: ERRORS:  -0.00255844  2.78368e-05 -7.28495e-07
[ WARN] [1734376965.412282038, 79.742000000]: ERRORS: -3.13879  -3.1412   3.1394
[ WARN] [1734376965.478943995, 79.790000000]: ERRORS: -3.13879  -3.1412  3.14159
[ WARN] [1734376965.549079151, 79.840000000]: ERRORS: -3.14129  3.13889 -3.14159
[ WARN] [1734376965.621986137, 79.890000000]: ERRORS: -3.14015  3.13775 -3.14159
[ WARN] [1734376965.691235946, 79.940000000]: ERRORS: -3.13993 -3.13842 -3.14159
[ WARN] [1734376965.762883946, 79.990000000]: ERRORS: -0.000475707  -0.00142249  2.37797e-07
[ WARN] [1734376965.829678191, 80.040000000]: ERRORS:  -0.00194066  -0.00124898 -1.18903e-07
[ WARN] [1734376965.901844773, 80.090000000]: ERRORS: -3.13928  3.13964 -3.14159
[ WARN] [1734376965.977797889, 80.140000000]: ERRORS: -0.00129048  0.00177276 4.60844e-08
[ WARN] [1734376966.054011594, 80.190000000]: ERRORS: -3.14046 -3.13916 -3.14159
[ WARN] [1734376966.130143541, 80.240000000]: ERRORS: -3.14149  -3.1369 -3.14159
[...]

PR #99, opened 2024-12-16
add heading control

talhabw:
We are using the integral of `cmd_vel.angular.z` to calculate a pose (target heading), then use it in our position pid controller to protect our heading from possible drifts. Best place to integrate this logic seems to be `depth_controller_node` because it already publishes `cmd_pose`. This is a draft for now, some tasks we need to do:

- [x] rename the node since it's no longer just a depth controller
- [x] remove some legacy code from the node (tau, current_depth, etc,)
- [x] decide if we are going to normalize our heading value? we don't need to (refer to #98 ) but we may want to?
- [x] test and tune it in the simulation and make sure it works as intended


senceryazici:
This will need to wait until #107 is merged and then will need to rebased, since CI is failing @talhabw

talhabw:
This worked fairly good on both the hardware and simulation tests. I think we can do the pid tuning part better, so merging this can wait until next test.

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+from auv_common_lib.control.enable_state import ControlEnableHandler
+
+
+class DepthControllerNode:
+    def __init__(self):
+        rospy.init_node("reference_pose_publisher_node")
+
+        # Initialize subscribers
+        self.odometry_sub = rospy.Subscriber(
+            "/taluy/odometry", Odometry, self.odometry_callback
```
senceryazici, line 17: "odometry"

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        self.set_depth_service = rospy.Service(
+            "/taluy/set_depth", SetDepth, self.target_depth_handler
```
senceryazici, line 20: "set_depth"

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        self.cmd_vel_sub = rospy.Subscriber(
+            "/taluy/cmd_vel", Twist, self.cmd_vel_callback
```
senceryazici, line 23: "cmd_vel"

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+
+import rospy
+from nav_msgs.msg import Odometry
+from geometry_msgs.msg import Pose, Twist
+from auv_msgs.srv import SetDepth, SetDepthRequest, SetDepthResponse
+from tf.transformations import quaternion_from_euler, euler_from_quaternion
+from auv_common_lib.control.enable_state import ControlEnableHandler
+
+
+class DepthControllerNode:
```
senceryazici, line 11: Rename this too

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self):
+        rospy.init_node("reference_pose_publisher_node")
```
senceryazici, line 13: We can move this to 

```py
if __name__ == "__main__"
    rospy.init....
```

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Initialize internal state
+        self.target_depth = 0.0
+        self.target_heading = 0.0
+        self.last_cmd_time = rospy.Time.now()
+
+        # Parameters
+        self.update_rate = rospy.get_param("~update_rate", 10)
+
+        rospy.Timer(rospy.Duration(1.0 / self.update_rate), self.control_loop)
```
senceryazici, line 39: I remember this implementation has its own problems when ROS is being shutdown. It was something that we encountered during the competition season, maybe @batuhanozer remembers

But the good practice is to simply remove the rospy.spin() in the run function and create the loop there using 

rate = ...
while not rospy.is_shutdown()
   ...
   rate.sleep()

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+            msg.pose.pose.orientation.w,
+        ]
+        _, _, self.target_heading = euler_from_quaternion(quaterion)
+
+    def cmd_vel_callback(self, msg):
+        if not self.control_enable_handler.is_enabled():
+            return
+
+        dt = (rospy.Time.now() - self.last_cmd_time).to_sec()
+        dt = min(dt, 0.1)
```
senceryazici, line 64: Can you move 0.1 to init as param?

This can be named
command_timeout, which later needs to be sync's with other nodes just like "control_rate" param.

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.cmd_pose_pub.publish(cmd_pose)
+
+    def run(self):
+        rospy.spin()
+
+
+if __name__ == "__main__":
+    try:
+        depth_controller_node = DepthControllerNode()
```
senceryazici, line 89: Again, name should be updated

auv_control/auv_control/launch/start.launch
```diff
@@ -36,7 +36,6 @@
     <node pkg="auv_control" type="depth_controller_node.py" name="depth_controller_node"
       output="screen">
       <param name="update_rate" value="$(arg rate)" />
```
senceryazici, line 38: I think you'll need the topic remappings. Even if the topics are 1:1 matched, we still do remap, example: from odometry to odmetry

Keep in mind the control enable handler subscribes to "control" topic too, so make sure you include that as well.

senceryazici (changes requested):
Good job, left some comments.

senceryazici (approved):
Looks good to me code-wise. Can be merged if the hardware test is working

ozanhakantunca (approved):

PR #105, opened 2024-12-21
WIP: smach shutdown handler

### LINKED ISSUE #104: Fix align frame controller still executing when smach is killed
Sometimes we kill the state machine mid task. At such times, some command velocity is still published by align frame controller. Not sure what the align frame controller tries to align to, maybe the frame it was trying to align itself before the smach is killed. But there were times I observed the robot moves at random and unexpected places, which might suggest the robot is trying to align somewhere else, or (x) there is some other bug causing the unwanted behaviour.

Whatever, the point is that when the smach is killed, we don't want to align anywhere apart from our current position. So we can either;
1. Call the disable service of align frame controller
2. Or align to taluy/base_link. Then there is no error, and cmd_vel is 0. 
 
I think the second option is more plausible. And this will probably solve the problem of unexpected behaviour (if x is not true)

talhabw:
Closes #104 

Sometimes we need to shutdown the smach to prevent the robot from getting damaged. This results in robot trying to align itself to the latest frame we have sent to `align_frame_controller`.

With this change we call the `taluy/control/align_frame/cancel` service when the smach is killed, which cancels the `align_frame_controller`, so the robot stays stable.

### DIFF
--- a/auv_smach/scripts/main_state_machine.py
+++ b/auv_smach/scripts/main_state_machine.py
@@ -3,7 +3,6 @@
 import rospy
 import smach
 import auv_smach
-from auv_smach.common import CancelAlignControllerState
 from auv_smach.initialize import InitializeState
 from auv_smach.gate import NavigateThroughGateState
 from auv_smach.red_buoy import RotateAroundBuoyState
--- a/auv_smach/scripts/main_state_machine.py
+++ b/auv_smach/scripts/main_state_machine.py
@@ -4,7 +4,6 @@ import rospy
 import smach
 import auv_smach
 from auv_smach.common import CancelAlignControllerState
-from std_srvs.srv import Trigger, TriggerRequest
 from auv_smach.initialize import InitializeState
 from auv_smach.gate import NavigateThroughGateState
 from auv_smach.red_buoy import RotateAroundBuoyState
--- a/auv_smach/scripts/main_state_machine.py
+++ b/auv_smach/scripts/main_state_machine.py
@@ -2,8 +2,8 @@
 
 import rospy
 import smach
-import signal
 import auv_smach
+from auv_smach.common import CancelAlignControllerState
 from std_srvs.srv import Trigger, TriggerRequest
 from auv_smach.initialize import InitializeState
 from auv_smach.gate import NavigateThroughGateState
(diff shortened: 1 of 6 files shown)


auv_smach/scripts/main_state_machine.py
```diff
@@ -42,6 +44,12 @@ def __init__(self):
 
         rospy.Subscriber("/taluy/propulsion_board/status", Bool, self.enabled_callback)
 
+        self.cancel_align_controller_service = rospy.ServiceProxy(
+            "/taluy/control/align_frame/cancel", Trigger
+        )
+
+        signal.signal(signal.SIGINT, self.shutdown_handler)
```
senceryazici, line 51: I believe the SIGINT signal already puts the main state machine in to preempted outcome, maybe when that outcome is achieved we just call the align frame controller and cancel it? Just asking if this signal handling is necessary

senceryazici (dismissed):

senceryazici:
This will need to wait until #107 is merged and then will need to rebased, since CI is failing @talhabw

auv_smach/scripts/main_state_machine.py
talhabw: SIGINT puts the main state machine to aborted outcome.

I think we can set each state's aborted state to  `CancelAlignControllerState` we already have. I think this may be a better way than signal handling.

senceryazici (dismissed):
maybe consider hooking up preempted to cancel align controller too. This comes down to the practice I guess, just see if this works out for your problem which you've described.

senceryazici (approved):
Looks good if all tested

talhabw:
this waits until #91 is merged

senceryazici:
What is the status here? after resolving the conflicts this can be merged no?

talhabw:
This didn't had the intended effects in hardware tests. We somehow couldn't catch the 'ctrl + c' exception and run the preempted/aborted states.

senceryazici:
> This didn't had the intended effects in hardware tests. We somehow couldn't catch the 'ctrl + c' exception and run the preempted/aborted states.

What steps need to be taken to finish this? Or is this obsolete now?

PR #107, opened 2024-12-22
temporarily remove dependencies requiring opencv dependencies

senceryazici:
The jetson uses it's own distributed opencv libraries therefore `libopencv-dev` must not be installed. Recently introduced changes enforce rosdep to install opencv libraries even though `--skip-keys` are provided. Removing the dependency for now. This will be later addressed.


batuhanozer (approved):
approved

PR #97, opened 2024-12-16
Create package structure for YOLO implementation

ozanhakantunca:
Closes #92 

The auv_vision meta-package and the auv_detection package have been created and properly configured. With this package structure, we can include our custom competition model without modifying third party repos and start the detection process.

PR #111, opened 2024-12-26
set default ultralytics device to 'cuda:0'

talhabw:
This fixes an error from #97 by setting default ultralytics device to `cuda:0` for jetson instead of cpu.

There is also a small addition to `.gitignore` for model files, i thought it was okay to put it in this pr since it's somewhat related.


senceryazici (approved):
LGTM

PR #112, opened 2024-12-26
add: follow path action server (A, merge to main)

eminmeydanoglu:
This pull request aims to introduce the carrot method (as in follow the carrot) to our navigation stack. Transitioning from our current direct alignment approach to a full plan-and-track method is a big task, requiring changes to how we handle a lot of things (looking at you, State Machine). Additionally, more sophisticated path planning algorithms than my straight line planner will be necessary for the future, for example for navigating around the buoy.

With this implementation, the main functionality for follow the carrot will be established. Given a target frame, we will be able to; generate a path to that frame, dynamically create a `carrot_frame` along the path, and track this carrot until the robot is aligned with the target. And we will still be able to align directly to a frame without planning. 

---
Here is a recording of me messing around with our new path following action server:
https://github.com/user-attachments/assets/d70281b7-8e7f-47ad-996d-fda9cb0828a3

1. I call the service to go to target with path planning on.
2. I test angle_offset and keep_orientation. These parameters change how the path is calculated, thus manipulating how the carrot orientation is calculated from path.
3. I do direct alignment. 
4. I do direct alignment with nonzero angle_offset and keep_orientation true.

All scenarios perform as expected. We also tested the path following part in our last pool test at 12.30.2024, and successfully validated the path-following functionality.

This PR introduces a significant number of changes. For my work to be understandable, I'll be making comments to respective parts and try to give my reasoning to why I did what I did.


auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,238 @@
...
+        self.server.start()
+        rospy.logdebug("FollowPath Action Server is ready.")
+        self.action_client.wait_for_server()
+        rospy.logdebug("Internal Action Client connected to the FollowPath Action Server.")
+        
+    def killswitch_callback(self, msg):
+        if not msg.data:
+            if self.server.is_active():
+                self.action_client.cancel_goal() 
+                rospy.logwarn("Control canceled")
```
eminmeydanoglu, line 74: This function will cancel the current goal and cause the robot to stop. Notice how we don't need to have a service to cancel the action as in legacy implementation of align frame controller. We can simply
```
rostopic pub /follow_path/cancel actionlib_msgs/GoalID "{}"
```
Action server publishes these topics:
/follow_path/cancel
/follow_path/feedback
/follow_path/goal
/follow_path/result
/follow_path/status

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+            
+        try:
+            feedback = FollowPathFeedback()
+            
+            while not rospy.is_shutdown(): # try to succeed until rospy is shutdown
+                
+                if self.server.is_preempt_requested():
+                    stop_cmd = Twist()
+                    self.cmd_vel_pub.publish(stop_cmd)
+                    self.server.set_preempted()
```
eminmeydanoglu, line 129: set the desired velocity state of controller to zero so that we stop.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+                        continue
+                    
+                    carrot_pose = calculate_carrot_pose(path, current_pose, self.carrot_distance)
+                    broadcast_carrot_frame(self.tf_broadcaster, self.tf_buffer, self.source_frame, carrot_pose)
+                    
+                    trans_error, rot_error = self.frame_controller.get_error(
+                        self.tf_buffer,
+                        self.source_frame,
+                        self.carrot_frame
+                        # We don't need to set angle_offset or keep_orientation here since they are taken cared of in path.
```
eminmeydanoglu, line 150: In this case, get error gets the error between robot and carrot. In direct alignment, we also add keep_orientation and angle_offset here, since we don't have the path anymore. Notice how the legacy align frame controller applied angle_offset when getting transform (get_transform is the ancestor of get_error):
```py
def get_transform(
        self, source_frame, target_frame, angle_offset, time=rospy.Time(0)
    ):
        try:
            # Get the current transform
            transform = self.tf_buffer.lookup_transform(
                target_frame, source_frame, time, rospy.Duration(2.0)
            )

            trans = (
                transform.transform.translation.x,
                transform.transform.translation.y,
                transform.transform.translation.z,
            )
            rot = (
                transform.transform.rotation.x,
                transform.transform.rotation.y,
                transform.transform.rotation.z,
                transform.transform.rotation.w,
            )

            return trans, rot

            # Apply the angle offset to the rotation
            roll, pitch, yaw = euler_from_quaternion(rot)
            yaw -= angle_offset  # Adjust yaw with angle offset
            yaw = angles.normalize_angle(yaw)
            new_rot = quaternion_from_euler(roll, pitch, yaw)
```

auv_navigation/auv_navigation/src/navigation_utils/align_frame_utils.py
```diff
@@ -0,0 +1,117 @@
...
+                transform.transform.rotation.z,
+                transform.transform.rotation.w,
+            )
+
+            _, _, yaw_error = euler_from_quaternion(rot)
+            
+            # Apply angle offset and handle keep orientation
+            yaw_error = self.wrap_angle(yaw_error + angle_offset)
+            if keep_orientation:
+                yaw_error = 0.0
```
eminmeydanoglu, line 53: We have this part for direct alignment, as I stated in https://github.com/itu-auv/auv-software/pull/112/files#r1900277444

auv_navigation/auv_navigation/src/navigation_utils/align_frame_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        
+        return twist
+    
+    def is_aligned(self, position_threshold, angle_threshold, current_pose,
+                    path, trans_error, rot_error):
+        if path is None:
+            total_trans_error = np.linalg.norm(trans_error)
+            within_position = (total_trans_error <= position_threshold)
+            within_angle = (abs(rot_error) <= angle_threshold)
+            return within_position and within_angle
```
eminmeydanoglu, line 81: If we are doing direct alignment, lets use our trans_error and yaw_error we got from get_error function. Since get_error function also takes care of keep_orientation and angle_offset, we can be sure that we are indeed checking our alignment to the exact pose we want to have.

auv_navigation/auv_navigation/src/navigation_utils/align_frame_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        
+        # For path-based alignment, calculate error between current pose and last path pose
+        last_pose = path.poses[-1].pose
```
eminmeydanoglu, line 84: We can just use the last pose of the path. Notice that when we are close to the end of the path, we calculate the carrot as exactly the same as the last path point. So, we can be sure that we are indeed checking our alignment to the exact pose of carrot at the end.

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+            
+            pose.pose.position.x = src_pos.x + t * (tgt_pos.x - src_pos.x)
+            pose.pose.position.y = src_pos.y + t * (tgt_pos.y - src_pos.y)
+            pose.pose.position.z = src_pos.z + t * (tgt_pos.z - src_pos.z)
+            
+            # Set orientation based on keep_orientation flag
+            if keep_orientation:
+                interp_q = src_q
+            else:
+                interp_q = quaternion_slerp(src_q, final_q, t)
```
eminmeydanoglu, line 62: We interpolate the yaw difference across the path. Any angle offset is added to the yaw difference.

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+            return carrot_pose
+        
+        remaining_distance -= segment_distance # Move to next segment
+        current_index += 1
+    
+    # If we've consumed all segments, return last pose
+    return path.poses[-1]
+
+
+def broadcast_carrot_frame(tf_broadcaster, tf_buffer, source_frame, carrot_pose):
```
eminmeydanoglu, line 168: I think having the carrot as a frame rather than a marker is more informative for debugging. Good thing about a frame is that we can also see its orientation, which is good to know. Also, I find having a carrot frame suits more to our path following implementation for several reasons.

eminmeydanoglu:
1. I call the service to go to target with path planning on.
2. I test angle_offset and keep_orientation. These parameters change how the path is calculated, thus manipulating how the carrot orientation is calculated from path.
3. I do direct alignment. 
4. I do direct alignment with nonzero angle_offset and keep_orientation true.

All scenarios perform as expected. We also tested the path following part in our last pool test at 12.30.2024, and successfully validated the path-following functionality.

https://github.com/user-attachments/assets/d70281b7-8e7f-47ad-996d-fda9cb0828a3

eminmeydanoglu:
Hey @senceryazici, this PR is now ready for a review. Looking forward to it!

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
senceryazici: also use following style to declare your functions
```py
def broadcast_carrot_frame(tf_broadcaster: tf2_ros.TransformBroadcaster, tf_buffer: tf2_ros.Buffer, source_frame: str, carrot_pose: geometry_msgs.msg.PoseStamped) -> None:
    ...

```

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        pose.header.stamp = transform.header.stamp 
+        pose.pose.position = transform.transform.translation
+        pose.pose.orientation = transform.transform.rotation
+        return pose
+    except (tf2_ros.LookupException, tf2_ros.ConnectivityException, 
+            tf2_ros.ExtrapolationException) as e:
+        rospy.logerr(f"Failed to get current pose: {e}")
+        return None
+
+def calculate_carrot_pose(path, robot_pose, carrot_distance):
```
senceryazici, line 96:
```py
def calculate_carrot_pose(path: nav_msgs.msg.Path, robot_pose: geometry_msgs.msg.PoseStamped, carrot_distance: float) -> geometry_msgs.msg.PoseStamped:


```

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        target_transform = tf_buffer.lookup_transform(
+            "odom", 
+            target_frame, 
+            rospy.Time(0),  
+            rospy.Duration(1.0)
+        )
+        src_pos = source_transform.transform.translation
+        tgt_pos = target_transform.transform.translation
+        
+        src_q = [source_transform.transform.rotation.x,
```
senceryazici, line 31: source_quaternion

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+            rospy.Time(0),  
+            rospy.Duration(1.0)
+        )
+        target_transform = tf_buffer.lookup_transform(
+            "odom", 
+            target_frame, 
+            rospy.Time(0),  
+            rospy.Duration(1.0)
+        )
+        src_pos = source_transform.transform.translation
```
senceryazici, line 28: source_position

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        tgt_pos = target_transform.transform.translation
```
senceryazici, line 29: target_position

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        segment_distance = np.linalg.norm(
+            np.array([
+                segment_end.position.x - segment_start.position.x,
+                segment_end.position.y - segment_start.position.y,
+                segment_end.position.z - segment_start.position.z
+            ])
+        )
+        
+        # Skip zero-length segments
+        if segment_distance < 1e-6:
```
senceryazici, line 136: move these static constants to the top of the file

```py
from ... import ...

# after imports
GOAL_DISTANCE_TOLERANCE=1e-6

```

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+
+import numpy as np
+import rospy
+from nav_msgs.msg import Path
+from geometry_msgs.msg import PoseStamped, TransformStamped
+import tf2_ros
+from tf.transformations import euler_from_quaternion, quaternion_from_euler, quaternion_matrix, quaternion_multiply, quaternion_inverse, quaternion_slerp
```
senceryazici, line 8: These can be confusing since you're using selective imports. Instead, use:


```py
tf.transformations.quaternion_multiply(..., ....)
````

This approach also improves readability aside from conflicting names. Longer function names don't necessarily reduce readability; in fact, they can make the code's purpose clearer.

File: auv_navigation/auv_navigation/setup.py [hunk 15]
```diff
@@ -0,0 +1,9 @@
+from setuptools import setup, find_packages
+from catkin_pkg.python_setup import generate_distutils_setup
+
+d = generate_distutils_setup(
+    packages=['navigation_utils'],
```
senceryazici, line 5: auv_navigation

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+            rospy.logwarn("FollowPath action did not succeed.")
+            if not self.server.is_preempt_requested():
+                self.server.set_aborted(result)
+
+        # Also publish stop command at the end of execute 
+        stop_cmd = Twist()
+        self.cmd_vel_pub.publish(stop_cmd)
+
+if __name__ == "__main__":
+    FollowPathActionServer()
```
senceryazici, line 232: move rospy.init_node to here.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+)
+from navigation_utils.align_frame_utils import AlignFrameController
+
+class FollowPathActionServer:
+    def __init__(self):
+        rospy.init_node('follow_path_action_server')
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+        self.tf_broadcaster = tf2_ros.TransformBroadcaster()
+        self.carrot_distance = rospy.get_param('~carrot_distance', 1.0)
```
senceryazici, line 26: is this the distance where the carrot is moved to the next pose? if so `dynamic_target_distance_threshold` is more appropriate?

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        self.position_threshold = rospy.get_param('~position_threshold', 0.1)
+        self.angle_threshold = rospy.get_param('~angle_threshold', 0.1)
+        self.source_frame = rospy.get_param('~source_frame', "taluy/base_link")
+        self.carrot_frame = rospy.get_param('~carrot_frame', "carrot")
```
senceryazici, line 30: dynamic_target_frame ?

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        self.control_rate = rospy.Rate(rospy.get_param('~control_rate', 20))
+        self.frame_controller = AlignFrameController(
+            max_linear_velocity=rospy.get_param('~max_linear_velocity', 0.8),
```
senceryazici, line 33:
```py

            max_linear_velocity=rospy.get_param('~align_to_frame_controller/max_linear_velocity', 0.8),
```

here and below

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+from navigation_utils.align_frame_utils import AlignFrameController
+
+class FollowPathActionServer:
+    def __init__(self):
+        rospy.init_node('follow_path_action_server')
+        self.tf_buffer = tf2_ros.Buffer()
+        self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
+        self.tf_broadcaster = tf2_ros.TransformBroadcaster()
+        self.carrot_distance = rospy.get_param('~carrot_distance', 1.0)
+        self.position_threshold = rospy.get_param('~position_threshold', 0.1)
```
senceryazici, line 27: `target_distance_threshold`

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        self.angle_threshold = rospy.get_param('~angle_threshold', 0.1)
+        self.source_frame = rospy.get_param('~source_frame', "taluy/base_link")
+        self.carrot_frame = rospy.get_param('~carrot_frame', "carrot")
+        self.control_rate = rospy.Rate(rospy.get_param('~control_rate', 20))
+        self.frame_controller = AlignFrameController(
+            max_linear_velocity=rospy.get_param('~max_linear_velocity', 0.8),
+            max_angular_velocity=rospy.get_param('~max_angular_velocity', 0.9),
+            kp=rospy.get_param('~kp', 0.55),
+            angle_kp=rospy.get_param('~angle_kp', 0.45)
```
senceryazici, line 36: instead of angle_kp, consider `angular_kp`

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+
+        self.path_pub = rospy.Publisher('path', Path, queue_size=1)
```
senceryazici, line 39: is this for visualization purposes?

senceryazici: if so, how about target_path?

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        self.cmd_vel_pub = rospy.Publisher('taluy/cmd_vel', Twist, queue_size=1) #TODO change to cmd_vel after given a namespace in start.launch
+        
+        self.killswitch_sub = rospy.Subscriber(
+            "/taluy/propulsion_board/status",
+            Bool,
+            self.killswitch_callback,
+        )
+        
+        rospy.Service(
+            "align_frame",
```
senceryazici, line 49: lets remove the service from this implementation completely.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
senceryazici: I like the idea however we may want to stop this action not only when killswitch is hit but also when battery is low, smach is preempted or killed, mission execution time is over etc etc.

We eventually will need to manage these services so this can be removed for now. the only enable/disable input we use for the whole control architecture is `control/enable` with type `std_msgs/Bool`, but I don't think that is necessary here since internal controllers will be off anyways.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+
+            return False # If exited the loop, success is False
+        except Exception as e:
+            rospy.logerr(f"Error during alignment: {e}") 
+            return False
+
+    def execute(self, goal):
+        rospy.loginfo("FollowPathActionServer: Received a new FollowPath goal.")
+        
+        if goal.path and goal.path.poses:
```
senceryazici, line 205: this approach is rather confusing. the action or server or the topic fields should not be populated selectively. if a secondary type of input is required, then you should have secondary topic (or action namespace) with it's own action message types.


 so either poses are provided 
 or 2 frames. I'm also in favor of separating that to another node which creates poses from the given 2 frames. the output of that node is connected to this node. etc etc

senceryazici (changes requested):
there seems to be a lot to discuss here. let's try to simplify the approach.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
eminmeydanoglu: I realized the parameter names can be confusing. Right now;
`carrot_distance` is the lookahead distance `calculate_carrot_pose` uses 
`position_thresold` and `angle_threshold` is used in `is_aligned` function. If current error is below those thresholds, `is_aligned` returns true. 
Lets change them to:
`dynamic_target_lookahead_distance`
`alignment_distance_tolerance`
`alignment_yaw_tolerance`

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
eminmeydanoglu: That sounds more appropriate. Maybe we will have a couple of active paths but only one will be the target_path.

eminmeydanoglu: I believe we can continue using `path` in the code, as it only cares about the target path if any.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -0,0 +1,233 @@
...
+        self.control_rate = rospy.Rate(rospy.get_param('~control_rate', 20))
+        self.frame_controller = AlignFrameController(
+            max_linear_velocity=rospy.get_param('~max_linear_velocity', 0.8),
+            max_angular_velocity=rospy.get_param('~max_angular_velocity', 0.9),
+            kp=rospy.get_param('~kp', 0.55),
+            angle_kp=rospy.get_param('~angle_kp', 0.45)
+        )
+
+        self.path_pub = rospy.Publisher('path', Path, queue_size=1)
+        self.cmd_vel_pub = rospy.Publisher('taluy/cmd_vel', Twist, queue_size=1) #TODO change to cmd_vel after given a namespace in start.launch
```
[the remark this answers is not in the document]
eminmeydanoglu: Alright. We can use `auv_navigation follow_path_action_server.py __ns:=taluy`until we prepare a launch file.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
eminmeydanoglu: Are you suggesting we should have a seperate node for the service?
I actually think that eventually a high level planner should completely take over service's job. That is, decide the target_frame, create path to target_frame (since it is a planner, it knows how to create the desired path to target whereas the service doesn't know), and start the action goal with the path. 
Until then, I think having a service creating path to target frame and starting goal is good practice. However I don't have the clearest picture. I think we can have a chat about the design for the overall system that will output goals for our action server.

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
eminmeydanoglu: in calculate_carrot_pose;
```py
carrot_pose.header = path.header
```
Path header is odom and we are calculating carrot in odom frame. We could just set carrot_pose header to be source frame here and not have to get source_frame as an argument in broadcast_carrot_frame, but then we would be lying about carrot's header. It would suggest that the carrot pose is defined in the source frame when it is actually in the odom frame.

auv_navigation/auv_navigation/src/navigation_utils/path_utils.py
eminmeydanoglu: Cool. I did a similar thing with 
```py
from navigation_utils.path_utils import (
    create_path_from_frame, get_current_pose, 
    calculate_dynamic_target_pose, broadcast_dynamic_target_frame
)
```
improves both clarity and maintainability.

eminmeydanoglu:
Oookay. Here is a list of updates and insights on the new design:
1. The action server is not a client to itself anymore. It simply receives a path and follows it. 
2. I made a couple of renaming and refactoring for clarity.
3. Lets talk about FrameAligner. I decided to have a `step()` function which gets error, calculates cmd_vel, and publishes, so action server's `do_path_following` method is less crowded. In `do_path_following`, we simply do:
 ```py
    self.frame_aligner.source_frame = self.source_frame
    self.frame_aligner.target_frame = self.dynamic_target_frame
    self.frame_aligner.angle_offset = 0.0
    self.frame_aligner.keep_orientation = False
    
    # perform one step of alignment loop
    self.frame_aligner.step() 
```
4. I thinked about what to do with align_frame_controller. The idea of directly using that controller node is more elegant, since then there will be only one node publishing cmd_vel, but comes with its own problems. Mainly, to use the controller node in our path tracking algorithm we will probably have to sacrifice from elegancy in other terms. Calling frame aligner's service in 20 hz is definitely not a good idea, and even though I came up with some other ideas other than using the service, none seemed easy and elegant.

   I thought that the best method was to seperate FrameAligner as a third party module, and use this module in the action server as above (and possibly in the align frame controller, but didn't implement that since it is not related to PR and since I don't really care about that node).

   And after all, we probably won't be needing the align frame controller node anyways. We also won't want a cmd_vel calculator but want a cmd_pose calculator. Here, I also thinked about changing our FrameAligner to a position controller that calculates cmd_pose but then again thought it is not related also, and figured it is not that easy to do that. 

   To sum up, in the future, we will only need the FrameAligner module (with a cmd_pose_calculator function) and our action server. I think having a FrameAligner module is elegant enough in that sense. 

With our path tracker in place, we can now shift our focus to developing the upper-level path planner. Combined that, we will achieve seamless and neat navigation along whatever desired path and crush the competition tasks :dancers:

**Comment by senceryazici** [comment]
I'll review this but first let's get the conflicts resolved and this rebased

**Comment by eminmeydanoglu** [comment]
> I'll review this but first let's get the conflicts resolved and this rebased

Oops. Done. I think those were some unnecessary conflict flag raises by Git.

**Comment by eminmeydanoglu** [comment]
Hey @senceryazici,
I’ve just added what I hope are the finishing touches to this PR by adding a launch file for the server node. As we discussed earlier today, feel free to take a look at the new changes and share your feedback.

**Comment by senceryazici** [comment]
This still needs to be rebased, and the CI must pass as well. pre commit checks fails as it seems

File: auv_msgs/CMakeLists.txt [hunk 26]
```diff
@@ -20,23 +28,23 @@ add_service_files(
   AlignFrameController.srv
   SetDepth.srv
   SetObjectTransform.srv
+  AlignFrame.srv
```
senceryazici, line 31: this needs to be removed

auv_msgs/action/FollowPath.action
```diff
@@ -0,0 +1,12 @@
+# Goal
+nav_msgs/Path path
+
+---
+# Result
+bool success
+float32 execution_time  # Time taken to execute the action in seconds
```
senceryazici, line 7: optional, use `std_msgs/Duration` here

auv_msgs/srv/AlignFrame.srv
```diff
@@ -0,0 +1,7 @@
+string target_frame
```
senceryazici, line 1: this file will be deleted

senceryazici (approved):

PR #114, opened 2024-12-29
Dynamic state machine execution

frk781:
Key Changes

- Enhanced state machine configuration
- Improved flexibility in mission state selection
- Simplified node initialization
- More robust parameter management

Configuration Updates
  - Added dynamic state mapping
  - Introduced test and full mission mode
  - Enabled easy state sequence modification
  - Improved launch file parameterization

To use the resent changes:
`roslaunch auv_smach start.launch`
-performes the whole competition states

 `roslaunch auv_smach start.launch test_mode:=true test_states:=init,buoy`
 -performes initialize and rotation around buoy 
 
can check the start.launch file for the shortend names of the states.

### DIFF
--- a/auv_smach/launch/start.launch
+++ b/auv_smach/launch/start.launch
@@ -1,7 +1,42 @@
+<?xml version="1.0"?>
 <launch>
-  <arg name="namespace" default="taluy" />
-  <param name="/smach/max_linear_velocity" value="0.3" />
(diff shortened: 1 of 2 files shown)


senceryazici:
Let's discuss about this. Maybe there is a way to simplify this even more

senceryazici (approved):
This seems to be useful for now, let's merge it

PR #117, opened 2025-01-08
WIP: Ozan/better logging

eminmeydanoglu:
1. Option to disable logging
We sometimes (actually in most of the times we launch start.launch) don't want the logging to start. Continuous logging during each launch can clutter the bag files, making it harder to search through them later and eating up storage space. To address this, I introduce a boolean flag to disable logging when not needed.
2. Logging functionality in simulation
Ability to log in simulation is super helpful for debugging! One can't keep track of all events and numbers whilst the simulation is running. Bug files enable a rigorous analysis of events that occurred during the runtime

### DIFF
--- a/auv_bringup/launch/start.launch
+++ b/auv_bringup/launch/start.launch
@@ -5,7 +5,7 @@
   <arg name="dvl_virtual_port" default="/tmp/vcom0" />
   <arg name="use_gui" default="false" />
   <arg name="logging_directory" default="$(optenv HOME)/bags" />
-
+  <arg name="enable_logging" default="true" />
   <rosparam file="$(find auv_bringup)/config/environment.yaml" command="load" ns="" />
 
   <!-- launch state publisher  -->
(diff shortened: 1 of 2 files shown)


ozanhakantunca:
I think that setting default="false" is much more reasonable. We test the robot very often, and not all of these tests are conducted in the pool. With default="true", we end up with a massive amount of bag records after every dry or pool test. Managing, filtering, and analyzing such a large volume of bag records becomes nearly impossible. I believe we can prevent this clutter by explicitly passing the argument during important moments when we actually need to record bags.

ozanhakantunca:
And also this PR needs to 
`pre-commit run --all-files`

eminmeydanoglu:
What if someone forgets to enable logging when we actually need it?

How about automatically starting logging whenever the DVL is active? That way we can ensure we are logging when the robot is in the water :d

The toggle can still remain available for cases where someone wants to start logging even when the DVL is inactive, or for some reason doesn't want to log when DVL is active

ozanhakantunca:
This seems like a very logical approach :D. Activating it when the DVL is operational appears to be the most appropriate option. However, I definitely don't think it's sustainable to rely on it recording a bag in every case.

senceryazici:
> How about automatically starting logging whenever the DVL is active

This could get you into trouble if DVL driver is not working as expected. It is best to not make crucial components rely on others. @eminmeydanoglu

senceryazici:
> I think that setting `default="false"` is much more reasonable

I disagree. By default it should be enabled, the critical setting is to turn this on instead off, since if set to `True` but produces unnecessary amount of bags, we can just delete it so it's just some amount of extra work. In other case where it needed to be enabled but it was not, you can never retrieve the data, then you have to re-do the test. Which is by far greater extra work IMO.

senceryazici:
How about if you had a script that cleans up all the log files less than a duration limit? rosbag has python lib that you can use, so a simple python script can inspect the given directory and if the bag's recording time is less than some threshold, just delete it.

ozanhakantunca:
Yeah, We tried eliminating bags smaller than 500 MB, but it seems there’s still a significant issue. Currently, during each test, approximately 5–10 bags larger than 1 GB are being generated.

I’m not sure if we need to use the DVL driver for this, but there should be a way to enable logging only when the robot is in the water. One potential approach could be linking the auv_localization_node service with the logging mechanism. Since we call this service consistently during every operation, it cannot be overlooked.

ozanhakantunca:
There is new PR that addresses logging issue here #157

PR #121, opened 2025-01-29
Add support for WrenchStamped

melihokur17:
Updated the controller_ros.h to publish WrenchStamped message type instead of Wrench to be able to subscribe in thruster_manager_ros.h and deal with transformation to the body frame and using it as is. Also tested in the simulation. This pr is updated version of old one #76 .


senceryazici (approved):
lets rebase first and merge

eminmeydanoglu:
Happy ending

PR #122, opened 2025-01-30
teleop xbox controller

melihokur17:
The purpose of this pr is that sometimes we have difficulty controlling the vehicle with the joystick in the tests, for instance, we want to move the vehicle only x or y axis, but when we do this with the joystick, we unintentionally make different movements. With this pr, as can be seen in the image, we will be able top drive these two axes with both the joystick part of the xbox controller and with the arrow keys, while we can drive in a single axis with the arrow keys indicated in red.

WhatsApp Image 2025-01-30 at 15 00 16

### DIFF
--- a/auv_teleop/config/xbox.yaml
+++ b/auv_teleop/config/xbox.yaml
@@ -1,0 +1,20 @@
+---
+buttons:
+  launch_torpedo1: 4
+  launch_torpedo2: 5
+  drop_ball: 3
+  z_control: 0
+
+axes:
+  x_axis:
+    index: [1, 7]
+    gain: 0.4
+  y_axis:
+    index: [0, 6]
+    gain: 0.4
+  z_axis:
+    index: [1, 7]
+    gain: 0.4
+  yaw_axis:
+    index: [3]
+    gain: 0.5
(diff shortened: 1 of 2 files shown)


ozanhakantunca (approved):

eminmeydanoglu:
I am thinking of repurposing the two upper buttons for z control (which currently trigger torpedo). I believe this is more convenient, and if I remember correctly we should never be triggering the torpedos right now (they activate go crazy mode) yet they look very easy to trigger mistakenly.

melihokur17:
I don't think the latest changes will affect using it with the logitech joystick, but I should still try it in the workshop. Since the places controlling z and x are different, the z_control property is not needed in this controller and has been removed. Here is the new layout.
WhatsApp Image 2025-02-06 at 00 24 36

auv_teleop/config/xbox.yaml
```diff
@@ -0,0 +1,22 @@
+---
+buttons:
+  launch_torpedo1: 2
+  launch_torpedo2: 1
+  drop_ball: 3
+  +z_axis:
```
senceryazici, line 6: Are these causing fails at formatting? perhaps `z_axis_pos` and `z_axis_neg` can be used

melihokur17: i tried the latest changes on both controllers in the simulation and there was no problem.

ozanhakantunca (approved):

PR #123, opened 2025-02-03
add birds eye camera to sim

melihokur17:
When we started the simulation with use_gui:= 0 to give our computer better performance, we needed to add a bird's eye view camera to the simulation because we thought it would be useful.


eminmeydanoglu:
Wow this is beyond helpful! Great idea :100: 
Screenshot from 2025-02-05 11-11-51

ozanhakantunca (approved):

PR #124, opened 2025-02-04
feature: react to `cmd_vel` commands in depth control

### LINKED ISSUE #115: Handling command timeout and desired state matrix in controller
Current implementation of controller_ros.h has a timeout function which returns 1 if there are no vel/pose commands for 1 second. And if timeout, it sends zero wrench;
```cpp
      const auto wrench_msg =
          (is_control_enabled() && !(is_vel_timeouted() && is_pose_timeouted()))
              ? auv::common::conversions::convert<ControllerBase::WrenchVector,
                                                  geometry_msgs::Wrench>(
                    control_output)
              : geometry_msgs::Wrench{};
```
There arises a couple of problems with this implementation if we are to move to a plan-and-control stack. On this pretext, we can also add some new cool functionalities:

### Problem: Desired states never reset even if timeouted.
Say we had cmd_pose z = -0.5 a time ago. But now we are not publishing it for 5 seconds. But since we had cmd_vel, controller didn’t think it is timeouted. So we continue to align with z=-0.5 even though we don’t want it anymore. 

This wouldn't be a problem before times, because we always have a cmd_pose z command, even when the depth controller service is never called (in that case cmd_pose z = 0)
But with our navigation stack, there will be times we don't want to have and cmd poses, instead we want all the control to be done by the velocity parts. 

But this would be simple to fix, right? Just imagine we seperated timeouted as pose is_vel_timeouted and is_pose_timeouted.

But not so easy. The question now is what to do if pose is timeouted or vel is timeouted. If both are timeouted, we can think about freezing the robot. That means
```cpp
desired_state_.head(6) = state_.head(6);
```
and
```cpp
desired_state_.tail(6).setZero();
```
So, the robot will freeze where it is timeouted for both pose and vel commands. Or we can just go with the old implementation, which is simply publish 0 wrench (I think freezing is better)
But what if we are only timeouted in one of them? Lets start with the easy one. If cmd vel is timeouted only, we will simply
```cpp
desired_state_.tail(6).setZero();
```
and position controller will do what it is supposed to.
But what if cmd pose is timeouted yet cmd vel active (which is our case for plan-and-control stack) ?

The problem here is that if we try to go with 

```cpp
desired_state_**.**head(6) **=** state_**.**head(6)**;**
```

the robot will want to both freeze and have a nonzero velocity. We obviously don’t want that.
Then we have to find a way to “ignore” desired_state position part if there are no pose commands. 

I will offer one solution I thought of, but maybe a better one can be found:

One possible solution is to set gains to zero temporarly. So something like this could work:
 ```cpp
if (is_pose_timeouted()) {
      // Update gains based on pose timeout
      update_gains_for_pose_timeout(is_pose_timeouted());
```
and
```cpp
void load_parameters() {
    kp_ = VectorRosparamParser::parse("kp", ros::NodeHandle("~"));
    ki_ = VectorRosparamParser::parse("ki", ros::NodeHandle("~"));
    kd_ = VectorRosparamParser::parse("kd", ros::NodeHandle("~"));
    
    // Store original gains
    kp_original_ = kp_;
    ki_original_ = ki_;
    kd_original_ = kd_;
  }

  void update_gains_for_pose_timeout(bool is_pose_timeouted) {
    auto controller = dynamic_cast<auv::control::SixDOFPIDController*>(controller_.get());
    if (!controller) return;

    if (is_pose_timeouted) {
      // Zero out position control gains (first 6 elements)
      kp_.head(6).setZero();
      ki_.head(6).setZero();
      kd_.head(6).setZero();
    } else {
      // Restore original position control gains
      kp_.head(6) = kp_original_.head(6);
      ki_.head(6) = ki_original_.head(6);
      kd_.head(6) = kd_original_.head(6);
    }

    controller->set_kp(kp_);
    controller->set_ki(ki_);
    controller->set_kd(kd_);
  }
```
This way, even though we will have a desired state slot from an old message (we can't seem to get rid of them) we won't be using it unless we are on velocity control. And if we are not also on velocity control, robot will just freeze (both commands are timeouted)

eminmeydanoglu:
Solves but not cl*ses #115 

This PR integrates `cmd_vel.linear.z` messages into `cmd_pose.linear.z` by accumulating velocity commands over time, just like how angular velocity commands were handled in [#99](https://github.com/itu-auv/auv-software/compare/emin/depth_control_cmd_vel?expand=1#99).

Also, since this is a small PR, and this second change is also about depth control, I changed the inital target_depth value to -0.4 from 0.0. This will prevent the robot to try to rush all the way up the moment we enable the controllers. Trust me, -0.4 is a carefully fine-tuned value! :)


auv_control/auv_control/scripts/align_frame_controller.py
```diff
@@ -140,7 +140,7 @@ def compute_cmd_vel(self, trans, rot):
         # Set linear velocities based on translation differences
         twist.linear.x = self.constrain(trans[0] * kp, self.max_linear_velocity)
         twist.linear.y = self.constrain(trans[1] * kp, self.max_linear_velocity)
-
+        twist.linear.z = self.constrain(trans[2] * kp, self.max_linear_velocity)
```
eminmeydanoglu, line 143: We can now calculate linear.z velocity commands

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -27,12 +27,12 @@ def __init__(self):
...
 
         # Initialize internal state
-        self.target_depth = 0.0
+        self.target_depth = -0.4
         self.target_heading = 0.0
         self.last_cmd_time = rospy.Time.now()
 
         # Parameters
-        self.update_rate = rospy.get_param("~update_rate", 10)
+        self.update_rate = rospy.Rate(rospy.get_param("~update_rate", 10))
```
eminmeydanoglu, line 35: Changed to type rospy.Rate instead of just storing an integer and converting it at run()

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -45,21 +45,21 @@ def odometry_callback(self, msg):
...
 
-        quaterion = [
+        quaternion = [
             msg.pose.pose.orientation.x,
             msg.pose.pose.orientation.y,
             msg.pose.pose.orientation.z,
             msg.pose.pose.orientation.w,
         ]
-        _, _, self.target_heading = euler_from_quaternion(quaterion)
+        _, _, self.target_heading = euler_from_quaternion(quaternion)
```
eminmeydanoglu, line 54: Fix a typo.

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -45,21 +45,21 @@ def odometry_callback(self, msg):
...
 
     def cmd_vel_callback(self, msg):
         if not self.control_enable_handler.is_enabled():
             return
 
         dt = (rospy.Time.now() - self.last_cmd_time).to_sec()
         dt = min(dt, self.command_timeout)
-
+        self.target_depth += msg.linear.z * dt
```
eminmeydanoglu, line 62: the magic happens here..

senceryazici (dismissed):
LGTM if this works as expected

eminmeydanoglu:
It works as expected and I think we should merge it.
Yet, I was doing some thought experiments, and figured it won't solve some edge cases, so we need other solutions (possibly masking). Actually let me write something about that later today to the issue.

eminmeydanoglu:
I will merge this once we migrate to the path following stack because (1) it introduces z axis velocity command computation which will cause problems with the current stack and (2) we don't need it till then anyways

eminmeydanoglu:
This PR is ready to go. @ozanhakantunca

auv_control/auv_control/scripts/reference_pose_publisher.py
ozanhakantunca, line 35: While both approaches work, storing the update rate as a number is logically sufficient. There's no need to wrap it immediately in a Rate object. And also, the first approach aligns more closely with ROS standards

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -78,11 +78,9 @@ def control_loop(self):
         self.cmd_pose_pub.publish(cmd_pose)
 
     def run(self):
-        rate = rospy.Rate(self.update_rate)
-
         while not rospy.is_shutdown():
             self.control_loop()
-            rate.sleep()
+            self.update_rate.sleep()
```
ozanhakantunca, line 83: Again, the first approach aligns more closely with ROS standards.

ozanhakantunca (changes requested):

Issue #126, opened 2025-02-04
Refactor straight_path_to_frame()


eminmeydanoglu:
There are a lot of things going on inside it. Lets seperate the inner methods as standalone functions, and combine all them in the main function.

PR #127, opened 2025-02-04
Emin: add: new path planning framework and path planner for gate task

eminmeydanoglu:
This PR adds a structured path planning framework, introducing the two new modules:

1. **`path_planners.py`**  
   -  the `PathPlanners` class for high-level path planning.  
   - Methods:  
     - **`straight_path_to_frame(...)`**: Generates a straight-line path between two TF frames. This function is very customizable to different scenarious like:
       - Select whether to interpolate position (X/Y, Z) and yaw  (if not, use the source values along the path)
       - Optional 360-degree rotations (`n_turns`)  
       - Additional `angle offset` (in radians) to add to the target orientation.
       This solves #126 by collecting all the helper functions to module 2.

     - **`path_for_gate(...)`**: Plans a two-segment path for the gate task.
       - From base link to the gate entrance  
       - From the entrance to the exit, adding a full turn.  

2. **`path_planning_utils.py`**  
   - Provides the `PathPlanningHelper` class. Details of functions can be found in code.

### DIFF
--- a/auv_navigation/auv_navigation/src/auv_navigation/path_planning_utils.py
+++ b/auv_navigation/auv_navigation/src/auv_navigation/path_planning_utils.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+import rospy
+import tf2_ros
+import tf
+import numpy as np
+from nav_msgs.msg import Path
+from geometry_msgs.msg import PoseStamped
+from typing import Optional
+
+DEFAULT_FIXED_FRAME: str = "odom"
+DEFAULT_FRAME_ID: str = "odom"
+TIME_ZERO = rospy.Time(0)
+TF_LOOKUP_TIMEOUT = rospy.Duration(1.0)
+TWO_PI: float = 2 * np.pi
+
+class PathPlanningHelper:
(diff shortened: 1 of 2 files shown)


auv_navigation/auv_navigation/src/auv_navigation/path_planning_utils.py
```diff
@@ -0,0 +1,174 @@
...
+DEFAULT_FIXED_FRAME: str = "odom"
+DEFAULT_FRAME_ID: str = "odom"
+TIME_ZERO = rospy.Time(0)
+TF_LOOKUP_TIMEOUT = rospy.Duration(1.0)
+TWO_PI: float = 2 * np.pi
+
+class PathPlanningHelper:
+    """Helper class containing static methods for path creation."""
+
+    @staticmethod
```
eminmeydanoglu, line 19: I made these methods static, so we don't have to instantiate this class when using it. However, we do have to instantiate path_planners.py since we have 
```py
def __init__(self, tf_buffer: tf2_ros.Buffer):
        self.tf_buffer = tf_buffer
```

Issue #128, opened 2025-02-04
bug: mission_start_link is almost identical to odom


eminmeydanoglu:
Lets explore initialize.py, the first state in our SMACH:

### 1. Resetting the Odometry Pose
```py
            smach.StateMachine.add(
                "SET_START_FRAME",
                SetStartFrameState(frame_name="mission_start_link"),
                transitions={
                    "succeeded": "SET_ALIGN_CONTROLLER_TARGET",
                    "preempted": "preempted",
                    "aborted": "aborted",
                },
            )
```
At the instance this state runs, odom and taluy_base_link becomes almost identical (with a pi/2 angle offset). and then over time, localization reintroduces the expected offset, gradually shifting the perceived base position. (at least thats my theory). How this looks in 3D will be introduced lately in the video below.

### 2. Creating the `mission_start_link` frame
```py
            smach.StateMachine.add(
                "SET_START_FRAME",
                SetStartFrameState(frame_name="mission_start_link"),
                transitions={
                    "succeeded": "SET_ALIGN_CONTROLLER_TARGET",
                    "preempted": "preempted",
                    "aborted": "aborted",
                },
            )
```
`SetStartFrameState` creates the frame `mission_start_link`, with a pose identical to `taluy/base_link` (since pose transformation is omitted) , and with the identity transformation applied
```py
transform_request.transform.transform.rotation.w = 1.0
```
so it is an instance of taluy/base_link right after odom is resetted.

Can you guess what happens when these two states are combined one after the other? Yes, the mission_start_link is basically the odom frame, with a pi/2 transformation in yaw applied. 

![Image](https://github.com/user-attachments/assets/08b8ea07-020d-4c5c-9af1-658a2587425f)

https://github.com/user-attachments/assets/2a4edbd6-24a4-481e-9e72-ae1317f3769c

Here, the real position of robot is z= 1.7 meters.

This is surely not desired since odom does not represent the taluy frame, especially right after it was resetted. 
And the third state is;
### 3. Alignment request to `mission_start_link` frame
```py
            # TODO add enable controller
            smach.StateMachine.add(
                "SET_ALIGN_CONTROLLER_TARGET",
                SetAlignControllerTargetState(
                    source_frame="taluy/base_link", target_frame="mission_start_link"
                ),
                transitions={
                    "succeeded": "succeeded",
                    "preempted": "preempted",
                    "aborted": "aborted",
                },
            )
```
[...]

PR #129, opened 2025-02-04
fix: fix alignment to mission_start_link

eminmeydanoglu:
This PR kinda addresses [#128  ](https://github.com/itu-auv/auv-software/issues/128)
For a detailed explanation, please do refer.

#### 1. Introduced the `DELAY_AFTER_RESET` state
After resetting the odometry, we now wait for one second before proceeding. This ensures mission_start_link does reflect the real pose of taluy.
For more details on why this approach works, see: [#128](https://github.com/itu-auv/auv-software/issues/128).

#### 2. Removed the alignment request for mission_start_link
We are requesting an immediate set_depth to go to gate depth. Why also try to align with mission_start_link and confuse the controllers?

#### 3. Publish `taluy/enable = "True"` in SetDepthState.
 This is needed now, since there is no ongoing alignment request while the state is running :)


auv_smach/src/auv_smach/common.py
```diff
@@ -27,6 +20,11 @@
     translation_from_matrix,
```
eminmeydanoglu, line 20: better imports

auv_smach/src/auv_smach/common.py
```diff
@@ -73,28 +71,67 @@ def concatenate_transforms(transform1, transform2):
...
         set_depth_request = SetDepthRequest()
         set_depth_request.target_depth = depth
         self.sleep_duration = sleep_duration
 
-        smach_ros.ServiceState.__init__(
-            self,
-            "/taluy/set_depth",
+        # Initialize the parent ServiceState
+        super(SetDepthState, self).__init__(
+            SET_DEPTH_SERVICE,
```
eminmeydanoglu, line 94: changed this to super(), so we avoid directly referring to the parent class. more flexible!

auv_smach/src/auv_smach/initialize.py
```diff
@@ -1,29 +1,20 @@
...
+from std_srvs.srv import SetBool, SetBoolRequest
+from std_srvs.srv import Empty, EmptyRequest
+from robot_localization.srv import SetPose, SetPoseRequest
+from auv_msgs.srv import SetObjectTransform, SetObjectTransformRequest
 from std_msgs.msg import Bool
-
-from auv_smach.common import SetAlignControllerTargetState, CancelAlignControllerState
+from auv_smach.common import CancelAlignControllerState
+from typing import Optional, Literal
+from dataclasses import dataclass
```
eminmeydanoglu, line 11: better imports

senceryazici:
The problem in #128 can potentially be solved with #135 . Can you try and see if that fixes the problem? If so we can avoid the extra work in here and keep it simple @eminmeydanoglu

eminmeydanoglu:
Apart from the problem you mentioned, this PR is still valid. It removes the align frame request to mission_start_link and publishes to control enable topic in backthread whilst at set depth state. And it doesn't hurt to have a 1 second delay for mission_start_link (It fixes a cruical bug for now anyways)

senceryazici:
This is a good temporary fix, but it does not address #128 properly.

the scope of this PR can be narrowed down to a temporary fix, where #128 will be addressed later on. @eminmeydanoglu

senceryazici:
Apart from that, looks good, again as temporary fix :D

senceryazici (approved):

auv_smach/src/auv_smach/common.py
```diff
@@ -27,6 +20,11 @@
     translation_from_matrix,
 )
 
+SET_DEPTH_SERVICE: str = "/taluy/set_depth"
```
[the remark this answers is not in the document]
eminmeydanoglu: This goes to #142

auv_smach/src/auv_smach/common.py
```diff
@@ -27,6 +20,11 @@
...
+SET_DEPTH_DEFAULT_SLEEP_DURATION: float = 5.0
+CONTROL_ENABLE_TOPIC: str = "/taluy/enable"
```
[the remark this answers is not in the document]
eminmeydanoglu: This goes to #142

PR #130, opened 2025-02-05
WIP: enhance: new gate smach with follow_path action server integration

eminmeydanoglu:
This PR improves the gate navigation state by utilizing the follow path action server for executing planned paths. Let me introduce the new players:

### 1. The action client: `follow_path_action_client.py`
It is time we introduce the action client for our server. It is pretty straighforward in the sense that the client is simply a python module which connects to the action server and transfers the given paths as goals. Then it returns the action server's result.
### 2. New State: `ExecutePlannedPathsState`
This new state in common module expects the input key "planned_paths". Then it calls the client function, and returns whatever client has returned (which returns whatever server has returned)
### 3. New State: `PlanGatePathsState`
This state uses the plan_for_gate() function (introduced in PR #127) to plan the path for gate task. Than it sends them as an input to `ExecutePlannedPathsState`

### DIFF
--- a/auv_smach/src/auv_smach/common.py
+++ b/auv_smach/src/auv_smach/common.py
@@ -1,24 +1,17 @@
 import smach
 import smach_ros
 import rospy
-from std_srvs.srv import SetBool, SetBoolRequest, SetBoolResponse
-from std_srvs.srv import Trigger, TriggerRequest, TriggerResponse
-from robot_localization.srv import SetPose, SetPoseRequest, SetPoseResponse
-from auv_msgs.srv import (
-    SetObjectTransform,
-    SetObjectTransformRequest,
-    SetObjectTransformResponse,
-    AlignFrameController,
-    AlignFrameControllerRequest,
-    AlignFrameControllerResponse,
-)
+import threading
+import numpy as np
+import tf2_ros
+import tf.transformations as transformations
+
+from std_srvs.srv import Trigger, TriggerRequest
+from auv_msgs.srv import AlignFrameController, AlignFrameControllerRequest
 from std_msgs.msg import Bool
 from geometry_msgs.msg import TransformStamped
-import tf2_ros
-import numpy as np
-import tf.transformations as transformations
-import tf2_geometry_msgs
-from auv_msgs.srv import SetDepth, SetDepthRequest, SetDepthResponse
+from auv_msgs.srv import SetDepth, SetDepthRequest
+from auv_navigation import follow_path_action_client
 
 from tf.transformations import (
     quaternion_matrix,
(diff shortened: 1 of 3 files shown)


auv_smach/src/auv_smach/common.py
```diff
@@ -1,24 +1,17 @@
...
 import tf.transformations as transformations
-import tf2_geometry_msgs
-from auv_msgs.srv import SetDepth, SetDepthRequest, SetDepthResponse
+
+from std_srvs.srv import Trigger, TriggerRequest
+from auv_msgs.srv import AlignFrameController, AlignFrameControllerRequest
+from std_msgs.msg import Bool
+from geometry_msgs.msg import TransformStamped
+from auv_msgs.srv import SetDepth, SetDepthRequest
+from auv_navigation import follow_path_action_client
```
eminmeydanoglu, line 14: better imports

auv_smach/src/auv_smach/gate.py
```diff
@@ -1,41 +1,51 @@
...
     SetAlignControllerTargetState,
     CancelAlignControllerState,
     SetDepthState,
+    ExecutePlannedPathsState
 )
 
-from geometry_msgs.msg import Twist
-from std_msgs.msg import Bool
 
-from auv_smach.initialize import DelayState
```
eminmeydanoglu, line 32: better imports

auv_smach/src/auv_smach/gate.py
```diff
@@ -45,61 +55,60 @@ def __init__(self, gate_depth: float):
         with self.state_machine:
             smach.StateMachine.add(
                 "SET_GATE_DEPTH",
-                SetDepthState(depth=gate_depth, sleep_duration=3.0),
+                SetDepthState(depth=gate_depth, sleep_duration=rospy.get_param("~set_depth_sleep_duration", 5.0)),
```
eminmeydanoglu, line 58: now we can set this parameter from launch file. Also, I calculated that 5 seconds is a better choice.

auv_smach/src/auv_smach/gate.py
```diff
@@ -45,61 +55,60 @@ def __init__(self, gate_depth: float):
...
-                "NAVIGATE_TO_GATE_EXIT",
-                NavigateToFrameState(
-                    "gate_enterance", "gate_exit", "gate_target", n_turns=-1
-                ),
+                "EXECUTE_GATE_PATHS",
+                ExecutePlannedPathsState(),
+                transitions={
+                    "succeeded": "CANCEL_ALIGN_CONTROLLER",
+                    "preempted": "CANCEL_ALIGN_CONTROLLER", # if aborted or preempted, cancel the alignment request
+                    "aborted": "CANCEL_ALIGN_CONTROLLER",  # to disable the controllers.
```
eminmeydanoglu, line 91: it is important we make sure the controller is cancelled always. If we abort at this state, the controllers will stay active and maybe bad things will happen.

PR #135, opened 2025-02-17
Use map frame for global orientation

senceryazici:
Can you try this? @eminmeydanoglu

### DIFF
--- a/auv_bringup/config/ekf.yaml
+++ b/auv_bringup/config/ekf.yaml
@@ -25,7 +25,7 @@ permit_corrected_publication: false
 map_frame: map
 odom_frame: odom
 base_link_frame: taluy/base_link
-world_frame: odom
+world_frame: map
 
 odom0: /taluy/localization/odom_imu
 odom0_config: [false, false, false,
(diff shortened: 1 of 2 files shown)


eminmeydanoglu:
I have been wondering around about this. Well it doesn't work out of the box. 
Here are some errors:
```
[ WARN]: Could not obtain transform from odom to map. Error was "map" passed to lookupTransform argument target_frame does not exist. 
[ERROR]: Could not obtain transform from odom->taluy/base_link. 
```
Looking at the tf tree, I notice the map frame is missing:
Screenshot from 2025-02-18 17-37-52
In Gazebo, the  `world`  frame is the global, fixed reference frame for the gazebo simulation. It is identical to odom, currently.
Notice how we don't have the frame `map`. And I am not sure if there is anything to broadcast it already. Which node is supposed to broadcast the map frame so map->odom exists? Was it supposed to come from the EKF part? Modifying the line in ekf.yaml doesn't seem to make that happen.
I tried to do the rename `map` -> `world`, that didn't seem to work out too. The transform odom -> base_link is still missing, and now I get:
```
[ERROR]: Could not obtain transform from odom to taluy/base_link. Error was Could not find a connection between 'taluy/base_link' and 'odom' because they are not part of the same tree.Tf has two or more unconnected trees.
```
I guess a way forward is not clear to me. 'd like to hear your thoughts. @senceryazici

PR #137, opened 2025-02-23
Update pressure odometry

### LINKED ISSUE #70: adjust sensor offset for bar30

melihokur17:
Closes #70 
Closes #62 
This is an updated version of pr #62 . It was a hassle to rebase it since it was a long time ago. It also caused problems with odometry, so I thought it would be more accurate to get the orientation information directly from the IMU sensor.

### DIFF
--- a/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
+++ b/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
@@ -90,9 +90,7 @@ class PressureToOdom:
         ]
 
         # Convert quaternion to 3x3 rotation matrix
-        rotation_matrix = tf.transformations.quaternion_matrix(quaternion)[
-            :3, :3
-        ]
+        rotation_matrix = tf.transformations.quaternion_matrix(quaternion)[:3, :3]
 
         sensor_offset = np.array(self.base_to_pressure_translation[0])
 
--- a/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
+++ b/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
@@ -49,7 +49,7 @@ class PressureToOdom:
 
         self.imu_data = None
         self.imu_subscriber = rospy.Subscriber(
-            "sensors/imu/data", Imu, self.imu_callback, tcp_nodelay=True
+            "imu/data", Imu, self.imu_callback, tcp_nodelay=True
         )
 
         self.depth_subscriber = rospy.Subscriber(
--- a/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
+++ b/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
@@ -74,11 +74,13 @@ class PressureToOdom:
     def get_base_to_pressure_height(self):
         try:
             if self.base_to_pressure_translation is None:
-                rospy.logerr_throttle(10,"Transform not available.")
+                rospy.logerr_throttle(10, "Transform not available.")
                 return 0.0
 
             if self.imu_data is None:
-                rospy.logwarn_throttle(10,"No IMU data received yet. Using default orientation.")
+                rospy.logwarn_throttle(
+                    10, "No IMU data received yet. Using default orientation."
+                )
                 return self.base_to_pressure_translation[0, 2]
 
             else:
@@ -103,7 +105,7 @@ class PressureToOdom:
                 return rotated_vector[2]
 
         except Exception as e:
-            rospy.logerr_throttle(10,f"Error in get_base_to_pressure_height: {e}")
+            rospy.logerr_throttle(10, f"Error in get_base_to_pressure_height: {e}")
             return 0.0
 
     def depth_callback(self, depth_msg):
--- a/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
+++ b/auv_navigation/auv_localization/scripts/pressure_odometry_node.py
@@ -16,6 +16,7 @@ import tf.transformations
 class PressureToOdom:
     def __init__(self):
         rospy.init_node("pressure_to_odom_node", anonymous=True)
+
         self.depth_calibration_offset = rospy.get_param(
             "sensors/external_pressure_sensor/depth_offset", 0.0
         )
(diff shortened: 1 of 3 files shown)


auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -44,19 +48,64 @@ def __init__(self):
...
         )
 
+    def imu_callback(self, imu_msg):
+        self.imu_data = imu_msg
+
+    def get_base_to_pressure_height(self):
+        try:
+            translation, _ = self.transformer.get_transform(
+                "taluy/base_link", "taluy/base_link/external_pressure_sensor_link"
+            )
```
eminmeydanoglu, line 68: the transform between taluy/base_link and the pressure sensor link is a static one. The value of it will never change—practically a constant. Yet in this implementation, we look up for this constant in every depth_callback call. This might add unnecessary computing?
 What if we look it up just once, maybe during __init__, and reuse it? So something like:
```py
    def __init__(self):
        # codes...
        self.static_translation = None
        try:
            """move the translation lookup here"""
        except Exception as e:
            rospy.logerr(f"Failed to lookup static transform")
```
and in get_base_to_pressure_height; simply 
 ```py
translation = self.translation
```

Comment (eminmeydanoglu) [reply auv_navigation/auv_localization/scripts/pressure_odometry_node.py:68]: Maybe a better solution is to put the node into a while loop looking up for the transform. So the transform doesn't have to be available at node init. And this while loop better not be inside init because of SoC

**Review by ozanhakantunca — COMMENTED** [review]
Transform from pressure sensor to base_link should be static

File: auv_navigation/auv_localization/scripts/pressure_odometry_node.py [hunk 2]
```diff
@@ -44,19 +48,73 @@ def __init__(self):
...
+                rate.sleep()
+
+    def imu_callback(self, imu_msg):
+        self.imu_data = imu_msg
+
+    def get_base_to_pressure_height(self):
+        try:
+            if self.base_to_pressure_translation is None:
+                rospy.logerr("Transform not available.")
+                return 0.0
```
eminmeydanoglu, line 80: If we are here, we definitely have the transformation. No need to check for it?

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -44,19 +48,73 @@ def __init__(self):
...
+
+            if self.imu_data is None:
+                rospy.logwarn("No IMU data received yet. Using default orientation.")
```
eminmeydanoglu, line 83: This will log pretty fast too.
rospy.logwarn_throttle(10, "No IMU data received yet. Using default orientation.")
Don't want eye fatigues!

eminmeydanoglu (approved):
LGTM

shaun-the-sheep-sheep

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -42,19 +46,75 @@ def __init__(self):
...
+            except Exception as e:
+                rospy.logwarn(f"Waiting for transform: {e}")
+                rate.sleep()
+
+    def imu_callback(self, imu_msg):
+        self.imu_data = imu_msg
+
+    def get_base_to_pressure_height(self):
+        try:
+            if self.base_to_pressure_translation is None:
```
senceryazici, line 76: if `self.base_to_pressure_translation` is guaranteed not to be `None` after while block in between [L62-L69](https://github.com/itu-auv/auv-software/pull/137/files#diff-035abc8e97639de0d6e8e1f354f4f2a57633ebd9d0fbdafecc958849ec465528R62-R69) then this check can be removed. even better this function expects the `self.base_to_pressure_translation` to be a valid number so an assert can be added

```py
def get_base_to_pressure_height(self):
    assert self.base_to_pressure_translation is not None
    # rest of the code
    
    try:
      

```

also do not place all the logic under the try block, only put the one that may throw the corresponding exception. so 



```py
def get_base_to_pressure_height(self):
    assert self.base_to_pressure_translation is not None
    if self.imu_data is None:
        rospy.logwarn_throttle(10, "No IMU data received yet. Using default orientation.")
        return self.base_to_pressure_translation[0, 2]

```

senceryazici: looking at it again, I don't think there is anything to be caught in a try except block? which line is expected to raise an exception here?

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -42,19 +46,75 @@ def __init__(self):
...
+                rospy.logerr_throttle(10, "Transform not available.")
+                return 0.0
+
+            if self.imu_data is None:
+                rospy.logwarn_throttle(
+                    10, "No IMU data received yet. Using default orientation."
+                )
+                return self.base_to_pressure_translation[0, 2]
+
+            else:
```
senceryazici, line 86: this `else` is not required since the above if block does an early return

**prefer this**
```py
def some_function(self):
    if some_condition:
        do_something()
        return something
   
    do_something_else()
    return something_else
```

**over this**
```py
def some_function(self):
    if some_condition:
        do_something()
        return something
    else:
       do_something_else()
       return something_else
```

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -42,19 +46,75 @@ def __init__(self):
...
+                sensor_offset = np.array(self.base_to_pressure_translation[0])
+
+                # Rotate the translation vector using the rotation matrix
+                rotated_vector = rotation_matrix.dot(sensor_offset)
+
+                return rotated_vector[2]
+
+        except Exception as e:
+            rospy.logerr_throttle(10, f"Error in get_base_to_pressure_height: {e}")
+            return 0.0
```
senceryazici, line 109: The `return 0.0` will be unnecessary after removing the try except block since nothing is expected to raise an exception here. 

But in general these kind of default returns must be used with caution since in this function it may cause the EKF to jump between values, which is something we don't want. We must handle all the cases we can, if not just let the `assert` raise an exception, or you manually raise one, which will stop the execution.

senceryazici (changes requested):

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -42,19 +47,65 @@ def __init__(self):
...
+    def imu_callback(self, imu_msg):
+        self.imu_data = imu_msg
+
+    def get_base_to_pressure_height(self):
+        assert self.base_to_pressure_translation is not None
+
+        if self.imu_data is None:
+            rospy.logwarn_throttle(
+                10, "No IMU data received yet. Using default orientation."
+            )
```
eminmeydanoglu, line 81: ### Quaternion sanity checks?
I was thinking about whether this check to imu data is all there is to consider. Notice that we are using the imu/data topic, which is the raw data we get directly from the sensor (and not the filtered odometry data). (And I actually don't see any reason to use the raw data instead of filtered odometry? But anyways now.)

We are using the imu data to create a rotation matrix. If the data is corrupted and we use it anyways, we will experience unexpected behaviours. And this data goes to odometry, which means it is pretty critical. 

I've come up with two possible checks that might be helpful. Check the below comment :beaver:

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
```diff
@@ -42,19 +47,65 @@ def __init__(self):
...
+            return self.base_to_pressure_translation[0, 2]
+
+        orientation = self.imu_data.orientation
+        quaternion = [
+            orientation.x,
+            orientation.y,
+            orientation.z,
+            orientation.w,
+        ]
+
```
eminmeydanoglu, line 91: What about adding...

A NaN check for the 4 elements of quaternion:
```py
# Check validity of imu quaternion data before using it for a transformation
        # Check for any NaN values
        if any(map(math.isnan, quaternion)):
            rospy.logwarn_throttle(
                10, "NaN values in IMU quaternion data. Using default orientation."
            )
            return self.base_to_pressure_translation[0, 2]
```

eminmeydanoglu: And a sanity check for whether imu data is normalized?
```py
norm = np.linalg.norm(quaternion)
if not np.isclose(norm, 1.0, atol=1e-3):
     # data is invalid. Quaternions should ideally have norm 1.
return self.base_to_pressure_translation[0, 2]

eminmeydanoglu: These checks ideally should be at imu_odometry_node (the node which takes in imu data and sends it to ekf). And again, I am not sure why we are not using the filtered orientation data and instead using raw imu. It seems we better implement these checks in the referred node so (1) we  won't send corrupted data to ekf and (2) we won't need to do these checks elsewhere, like in this node.

ozanhakantunca: I believe it would be more beneficial to address the orientation data validation in a separate PR on imu_odometry_node later, as this PR has already been waiting for quite some time. Therefore, let's focus on evaluating the functionality of the current PR for now. I recall that in the early stages of this PR, we started by taking IMU data from odometry. However, for reasons I can't quite recall, we decided that it would be more appropriate to use raw data directly from the sensor. We can still discuss this matter further.

ozanhakantunca (approved):
It looks good to me overall :D But we can wait until the next meetup to discuss the basis of the orientation data before merging.

auv_navigation/auv_localization/scripts/pressure_odometry_node.py
eminmeydanoglu: Yeah that makes sense. I am curious about why odometry quaternion data didn't work out. I see the node ends up publishing what it subscribes to, but that doesn't seem too problematic 😂

melihokur17: I now tried getting orientation data from the topic /localization/odom_imu and there doesn't seem to be any problems.

ozanhakantunca:
We discussed it within the team and resolved all concerns. It's ready to be merged in its current state, but it still requires approval. @senceryazici

eminmeydanoglu:
image

PR #138, opened 2025-02-23
auv_cam pkg into auv_hardware 

melihokur17:
Moved auv_cam package into auv_hardware_bridge package.


auv_hardware/auv_hardware_bridge/CMakeLists.txt
```diff
@@ -1,18 +1,45 @@
...
+    opencv_core
+    opencv_imgproc
+    opencv_imgcodecs
+    opencv_videoio
+  CONFIG
+)
+
 catkin_package(
   INCLUDE_DIRS include
+#  LIBRARIES auv_cam
```
senceryazici, line 40: remove auv_cam

senceryazici:
auv_hardware_bridge is a dependency of auv_hardware, which should be dependency of auv_software

senceryazici (approved):

senceryazici:
also add
```xml
  <depend>auv_hardware</depend>
```

to package.xml under auv_software

ozanhakantunca (approved):
I didn't notice any mistakes here, but @senceryazici's review might also be helpful.

ozanhakantunca (changes requested):
I had a discussion with @senceryazici, and he recommended that this approach would be more suitable. We should place the auv_cam package directly under auv_hardware as a complete package since we are using the auv_hardware_bridge package as a protocol bridge. As a result, we will move the auv_cam package under auv_hardware without modifying its contents, and subsequently remove it from the auv_software dependency tree and add it to the auv_hardware dependency tree. The updated package structure should look like this:

```
   auv_hardware/
      -auv_hardware
      -auv_hardware_bridge
      -auv_cam
```

ozanhakantunca (approved):

PR #143, opened 2025-03-01
WIP: Create roll compensation module

ozanhakantunca:


ozanhakantunca:
Screenshot from 2025-03-03 14-21-36

I simulated a small scene in the image. We can see that YOLO detection works properly with the dynamic_roll_compensation_node. At that moment in the image, /taluy/odometry reported a roll angle of 0.13 radian (23.14 degrees).

That is, even at nearly the maximum roll angles we might encounter, the shrinking of the image does not pose a significant problem for us, in my opinion.

I have a few questions regarding whether the frequent corrections applied to the image during an actual pool test might cause noticeable jitter. Therefore, I believe it is essential to include this in the hardware test.

ozanhakantunca:
There is a CI bug in this PR since I added the launch files to simulation bringup. I will try the solve but if it takes too long we can merge this without launch configuration in simulation environment.

ozanhakantunca:
Hey @senceryazici  😄 This PR has been waiting for a few months due to a CI failure I can’t resolve. We even tried matching the OpenCV dependency setup from `auv_hardware`, but it still fails. Could you take a look when you have a moment?

PR #146, opened 2025-03-03
Add taluy gui

melihokur17:


ozanhakantunca:
What is the current situation here @melihokur17 ?

auv_teleop/scripts/taluy_gui.py
```diff
@@ -0,0 +1,179 @@
...
+        self.pool_window.show()
+
+    def open_simulation_mode(self):
+        self.hide()
+        self.simulation_window = MainControlPanel(include_simulation=True)
+        self.simulation_window.show()
+
+
+if __name__ == "__main__":
+    rospy.init_node("taluy_gui_node", anonymous=True)
```
eminmeydanoglu, line 175: One incovenience I see is, pressing the X button won't stop the processes taluy gui started. For example, any topic publishing active will stay active.
For an easy (though not a complete) solution, I think we can modify this code such that the node will be killed when we press the X, thus no more zombie publishing from taluy gui.

Not sure the exact way to achieve this. Here is an AI suggestion:
```py
if __name__ == "__main__":
    # Import signal module for graceful shutdown
    import signal

    # Initialize ROS node first
    # anonymous=True ensures unique node name if multiple instances are run
    rospy.init_node("taluy_gui_node", anonymous=True)

    app = QApplication(sys.argv)

    # Handle Ctrl+C (SIGINT) gracefully by setting the default handler.
    # This allows the Python interpreter to raise KeyboardInterrupt,
    # which Qt's event loop can then handle, eventually leading to
    # the closeEvent being triggered if the main window is active.
    signal.signal(signal.SIGINT, signal.SIG_DFL)

    # Create and show the main application window
    window = StartScreen()
    window.show()

    # Start the Qt event loop. Execution blocks here until the application exits.
    exit_code = app.exec_()

    # Ensure ROS node shuts down cleanly after Qt application exits
    # This is important to unregister publishers/subscribers etc.
    rospy.signal_shutdown("Qt application exited")

    # Exit the script with the same code returned by app.exec_()
    sys.exit(exit_code)

**Review by eminmeydanoglu — APPROVED** [review]
Sorry for the late review.

I left some suggestions and concerns, though I don't think they all need to be addressed before merging, since this code is not a part of the autonomous system.

**Comment by eminmeydanoglu** [comment]
One more comment:
I don't see why we have the two tabs "pool" and "simulation".
The pool tab is missing the SMACH block. Isn't it supposed to be there?
Why not unify the two tabs, since all the service/ topic names are already the same. From the gui's perspective, there is no simulation / real life distinction. 

Again, this isn't urgent to address.
Screenshot from 2025-04-06 17-09-45
Screenshot from 2025-04-06 17-08-33

**Review by ozanhakantunca — CHANGES_REQUESTED** [review]
Sorry for the late review. This is a big PR with lots of work to done. I think @eminmeydanoglu's review is a great starting point. Let’s take another look after the requested changes are submitted.

Technically, we could merge this now, but since there’s no urgency and I believe it’s important to keep the codebase in main as clean and well-organized as possible, I’d prefer to hold off a bit and go through one more review cycle.

File: auv_teleop/scripts/vehicle_control_tab.py [hunk 2]
```diff
@@ -0,0 +1,216 @@
...
+        if not self.enable_publishing:
+            self.enable_publishing = True
+            self.enable_thread = threading.Thread(target=self.publish_enable)
+            self.enable_thread.start()
+
+    def stop_enable_publishing(self):
+        self.enable_publishing = False
+        if self.enable_thread and self.enable_thread.is_alive():
+            self.enable_thread.join()
+        self.enable_pub.publish(False)
```
ozanhakantunca, line 216: Maybe you could use a more descriptive name. It’s hard to understand the purpose or specifics of the function just from its current name.

ozanhakantunca (changes requested):
Looks good to me overall, just a few minor things left out 😆 @melihokur17

auv_teleop/package.xml
```diff
@@ -7,18 +7,30 @@
...
+  <build_export_depend>std_msgs</build_export_depend>
+  <build_export_depend>std_srvs</build_export_depend>
+  <exec_depend>auv_msgs</exec_depend>
   <exec_depend>geometry_msgs</exec_depend>
   <exec_depend>joy</exec_depend>
+  <exec_depend>message_runtime</exec_depend>
   <exec_depend>rospy</exec_depend>
   <exec_depend>sensor_msgs</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+  <exec_depend>std_srvs</exec_depend>
```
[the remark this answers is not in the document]
eminmeydanoglu: I guess you are talking about using single <depend> tag for the package :)

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+            try:
+                self.launch_process.wait(timeout=2)
+            except subprocess.TimeoutExpired:
+                print("Process did not terminate, killing it...")
+                self.launch_process.kill()
+            self.launch_process = None
+        else:
+            print("No teleop process to stop.")
+
+    def publish_velocity(self):
```
eminmeydanoglu, line 164: publish_velocity_loop? This function is intended to run continuously
```suggestion
    def publish_velocity_loop(self):
```

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+        rate = rospy.Rate(20)
+        while self.publishing and not rospy.is_shutdown():
+            self.cmd_vel_pub.publish(self.current_twist)
+            rate.sleep()
+
+    def start_publishing(self, direction, speed):
```
eminmeydanoglu, line 170: Well this doesn't "start" a thread right? It checks the manual input commands and sets current_twist. What about update_manual_velocity? 
```suggestion
    def update_manual_velocity(self, direction, speed):
```

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+        if action:
+            action()
+        else:
+            print(f"Unknown direction: {direction}")
+
+        if not hasattr(self, "publish_thread") or not self.publish_thread.is_alive():
+            self.publish_thread = threading.Thread(target=self.publish_velocity)
+            self.publish_thread.start()
+
+    def stop_publishing(self):
```
eminmeydanoglu, line 195: This resets current_twist to zero. reset_manual_velocity? 
```suggestion
    def reset_manual_velocity(self):
```

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+        self.publishing = False
+        self.current_twist = Twist()
+        self.cmd_vel_pub.publish(self.current_twist)
+
+    def publish_enable(self):
```
eminmeydanoglu, line 200: Again lets add _loop to clarify its role as a running thread. And make sure everybody understands its about enabling control. And _signal will make it clear its a boolean on-off signal to enable/disable controllers
```suggestion
    def publish_control_enable_signal_loop(self):
```

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+        rate = rospy.Rate(20)
+        while self.enable_publishing and not rospy.is_shutdown():
+            self.enable_pub.publish(True)
+            rate.sleep()
+
+    def start_enable_publishing(self):
```
eminmeydanoglu, line 206: Using the above reasoning, 

```suggestion
    def start_control_enable_signal_publishing(self)
```

auv_teleop/scripts/vehicle_control_tab.py
```diff
@@ -0,0 +1,216 @@
...
+        if not self.enable_publishing:
+            self.enable_publishing = True
+            self.enable_thread = threading.Thread(target=self.publish_enable)
+            self.enable_thread.start()
+
+    def stop_enable_publishing(self):
```
eminmeydanoglu, line 212:
```suggestion
    def stop_control_enable_signal_publishing(self)  
```

eminmeydanoglu (commented):
I left some naming suggestions for vehicle_control_tab.py.

auv_teleop/scripts/services_tab.py
```diff
@@ -202,6 +223,13 @@ def enable_dvl(self):
         else:
             QMessageBox.warning(self, "Error", "Failed to enable DVL.")
 
+    def clear_objects(self):
```
ozanhakantunca, line 228: We can rename this func to clear_object_transforms later

ozanhakantunca (approved):
Overall, it looks good. Given that quite a bit of time has passed, I think we can merge this as is. We can continue improving it after it’s merged into main.

PR #157, opened 2025-03-07
feat: add logging configuration options to launch files

ozanhakantunca:
This PR is updated version of this PR (#117) . Because of the long journey that we cannot fix the issue in the logging functionality, I just added the arguments that open or disable logging for preventing the huge mass of bag files from the dry tests.


eminmeydanoglu:
LGTM apart from the change in logging.launch. It doesn't seem related and couldnt get the purpose.

ozanhakantunca:
Yes, as I already mentioned, instead of trying to integrate the auv_logger system, introduced a temporary solution using the logging_enable argument. Since we have much more critical issues to focus on right now, we can prevent the current problem with this temporary fix and revisit it later when appropriate.

PR #162, opened 2025-03-17
Fix entrance and exit frames

### LINKED ISSUE #119: Fix gate entrance and exit frame positions
The z axis positions of gate entrance and exit frames are in the same level with the arrow frames. This creates a misleading representation of the precise entry and exit points for the gate. To address this, we should incorporate the gate_depth parameter (currently defined in SMACH) when calculating these frames. 

Also, lets create those frames relative to gate links!

ozanhakantunca:
Closes #119


eminmeydanoglu (approved):
Almost LGTM. Leaving some suggestions... 
I approve the PR (+ my suggestions added).

Just a reminder: This node will likely evolve into a "offset calculator" that computes two points perpendicular to gate line. And we will just feed them to navigation. 

When that day comes, we won't need to broadcast transforms or try to determine whichever is gate_entrance (which you removed already and is fine :D).

melihokur17 (approved):

PR #163, opened 2025-03-17
Various refactoring to auv_navigation

eminmeydanoglu:


eminmeydanoglu:
I believe this simple PR is clear enough for our team members to review and merge.  
@ozanhakantunca @talhabw @melihokur17 @frk781 @Temponer, your reviews are appreciated.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
```diff
@@ -11,7 +11,7 @@
     FollowPathResult,
     FollowPathActionGoal,
 )
-from auv_navigation import path_utils
+from auv_navigation.follow_path_action import follow_path_helpers
```
ozanhakantunca, line 14: Is follow_path_helpers differ from path_utils or they represents the same thing? If they are same, I believe that "path_utils" is a better naming. IMO :D

ozanhakantunca (changes requested):
Leaving small changes but LGTM at all. We can merge this after solving the discussions. Thanks for the corrections @eminmeydanoglu :D.

auv_navigation/auv_navigation/scripts/follow_path_action_server.py
eminmeydanoglu: This is a rename, and *_utils.py is never a good naming and suggest code smells. Check this out: https://chatgpt.com/share/67df30a0-0000-8007-af61-6b1a316bcb0e

Also, I think it is reasonable to say that we should seperate this helper module to different submodules, for example dynamic_target_helpers.py and transformations.py (for transform related functions like get_robot_pose). But let that be the topic of another PR.

auv_navigation/auv_navigation/package.xml
```diff
@@ -7,6 +7,10 @@
   <author email="senceryazici@gmail.com">Sencer Yazici</author>
   <license>BSD-3-Clause</license>
   <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>actionlib</build_depend>
```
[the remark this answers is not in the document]
eminmeydanoglu: Cool. Thanks.

PR #169, opened 2025-03-24
Meydanoglu/orientation handling for dynamic target

eminmeydanoglu:


eminmeydanoglu:
### Update

This project is harder than I imagined. I've created this function for a solution:
```py
def select_dynamic_target(
    robot_pose: PoseStamped,
    candidate_dynamic_target: PoseStamped,
    dynamic_target_yaw_threshold: float,
    last_dynamic_target: Optional[PoseStamped],
) -> PoseStamped:
    """
    Determine the dynamic target to use based on yaw difference.
    1. If yaw_diff <= threshold, update and return the candidate.
    2. If yaw_diff > threshold:
        a. If there is a last dynamic target, return it (freeze).
        b. If not, use the candidate as the first dynamic target and freeze.
    """
    _, _, robot_yaw = tf.transformations.euler_from_quaternion(
        (
            robot_pose.pose.orientation.x,
            robot_pose.pose.orientation.y,
            robot_pose.pose.orientation.z,
            robot_pose.pose.orientation.w,
        )
    )
    _, _, candidate_yaw = tf.transformations.euler_from_quaternion(
        (
            candidate_dynamic_target.pose.orientation.x,
            candidate_dynamic_target.pose.orientation.y,
            candidate_dynamic_target.pose.orientation.z,
            candidate_dynamic_target.pose.orientation.w,
        )
    )
    
    # absolute and shortest angular distance
    yaw_diff = abs(angles.shortest_angular_distance(0.0, robot_yaw - candidate_yaw))

    if abs(yaw_diff) > dynamic_target_yaw_threshold and last_dynamic_target is not None:
        rospy.loginfo(
            "Freezing dynamic target. Candidate yaw diff: {:.2f} rad.".format(yaw_diff)
        )
        return last_dynamic_target
    msg = (
        "Using candidate dynamic target."
        if abs(yaw_diff) <= dynamic_target_yaw_threshold
        else "First candidate is freezable. Using it as dynamic target."
    )
    rospy.loginfo("{} Yaw diff: {:.2f} rad.".format(msg, yaw_diff))
    # if yaw_diff < threshold, update and return the candidate
    # if yaw_diff >= threshold but last_dynamic_target is None, use candidate as first dynamic target
    return candidate_dynamic_target
```
This function was meant for a validity check for the candidate (computed) dynamic target. 

Well, it is very buggy. Namely, think about the scenario where taluy and dynamic frame becomes identical, and the computed candidate is freezable (and we freeze everything freezable). In this situation, we get stuck indefinitely, since the dynamic target only updates if the candidate is considered valid, which it no longer is in this frozen state.

To address this, I’m exploring a more robust solution. My idea is to introduce a cost minimizing framework that will find the next best dynamic target by perfecting the two factors: maintaining a desirable lookahead distance and respecting yaw threshold constraints. I will be working on achieving this next.

PR #170, opened 2025-04-02
Several Changes to KF node

frk781:
PR Summary
    Distance-based Indexing Addition: The `object_server` now includes **_distance-based indexing_** based on the taluy/_base_link frame_. After discussion with the team, we agreed that calculating the distance is sufficient for indexing objects with the same name in different lanes. We **decided against** using exclusion methods like drawing lines from **gate arrows** to determine the left or right side.

Camera Detection Pose Estimator Refactor  A minor refactor was made in the camera_detection_pose_estimator node to address rotation. Since the camera's optical link's  axes differ from the odom frame, the necessary transformation was added (manually). Rotation is not actually important or used currently. I just had to do this because all calculations in the repo assume that objects have the same rotation relative to the odometry frame.

`#145` Changes Implemented: Changes requested in pr #145(seperation of `prop_transform_publisher`) regarding the camera_detection_pose_estimator were also implemented.
    
   **`SIDE NOTE`**: this changes (Kalman Filter additon) are tested both in pool and simulation. We decided to use kalman filter instead of the old EMA filter method beacause we found results  were better with kf.

### DIFF
--- a/auv_navigation/auv_mapping/include/auv_mapping/object_map_tf_server_ros.hpp
+++ b/auv_navigation/auv_mapping/include/auv_mapping/object_map_tf_server_ros.hpp
@@ -38,7 +38,7 @@ class ObjectMapTFServerROS {
 
   void dynamic_transform_callback(
       const geometry_msgs::TransformStamped::ConstPtr &msg);
-
+  void update_filter_frame_ids(const std::string &object_frame);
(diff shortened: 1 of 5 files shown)


ozanhakantunca (approved):
Sorry for the late review, looks good to me overall.
We can do a more detailed review before merging it into main.

PR #176, opened 2025-04-14
Improved SetFrameLookingAtState

melihokur17:


eminmeydanoglu:
Hey @melihokur17, check out the issue #177 

Will this PR address everything there? And maybe the structure I tried to draw will help

melihokur17:
Yep, this includes everything I aimed for in this pr. Thanks for this well written issue.

eminmeydanoglu:
Excited to review this.

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
             return "aborted"
 
 
+class RotationState(smach.State):
+    def __init__(
```
eminmeydanoglu, line 308:
```suggestion
    """Rotates the AUV until a TF transform between `source_frame` and `look_at_frame`
    becomes available, or until a full 360-degree rotation is completed if
    `full_rotation` is True.
    """
```

melihokur17: 1. full_rotation = True (Full Rotation):
succeeded (Success):
- The robot completes a full 360-degree rotation even if it finds the transform.
- After the full rotation, if the transform is found, the state succeeds.

aborted (Failure):
- If the transform is not found after the full rotation, the state aborts.

2. full_rotation = False (No Full Rotation):
succeeded (Success):
- If the transform is already available before rotation starts, the state succeeds immediately.
- If the transform is found during the rotation, the state succeeds and stops rotating.

aborted (Failure):
- If the transform is not found until completing a full rotation, the state aborts.

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
...
+        else:
+            rospy.loginfo(
+                "RotationState: completed full rotation but no transform found between %s and %s",
+                self.source_frame,
+                self.look_at_frame,
+            )
+            return "aborted"
+
+
+class SetFrameLookingAtState(smach.State):
```
eminmeydanoglu, line 464: I find this name very confusing. It is not clear what we set (the class doesnt seem to set anything). And what does it mean "Looking at frame" ? I doubt any class name should include a word that ends with -ing.

A suggestion: BroadcastOrientToObjectFrame. This will signal that "This state will broadcast a TF frame that is oriented towards an object."

Or instead of object, we can use target.

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
...
+    def __init__(self, source_frame, look_at_frame, target_frame, duration_time=3.0):
+        smach.State.__init__(self, outcomes=["succeeded", "preempted", "aborted"])
+        self.source_frame = source_frame
+        self.look_at_frame = look_at_frame
+        self.target_frame = target_frame
```
eminmeydanoglu, line 469: So this target_frame is the newly created frame, and we ask the align frame to align with it. But the word target is confusing in this context. 

What about "alignment_frame"
"new_pointed_frame"

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
...
+                self.look_at_frame,
+            )
+            return "aborted"
+
+
+class SetFrameLookingAtState(smach.State):
+    def __init__(self, source_frame, look_at_frame, target_frame, duration_time=3.0):
+        smach.State.__init__(self, outcomes=["succeeded", "preempted", "aborted"])
+        self.source_frame = source_frame
+        self.look_at_frame = look_at_frame
```
eminmeydanoglu, line 468: I think look_at is okay in terms of naming. Though I also have the ideas:
aiming_target_frame
object_target_frame
target_frame (If there is a target anywhere, it is this one :))

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
...
+                )
+
+                direction_vector = np.array(
+                    [
+                        base_to_look_at_transform.transform.translation.x,
+                        base_to_look_at_transform.transform.translation.y,
+                    ]
+                )
+
+                facing_angle = np.arctan2(direction_vector[1], direction_vector[0])
```
eminmeydanoglu, line 504: target_yaw_angle_rad

auv_smach/src/auv_smach/common.py
```diff
@@ -288,6 +304,250 @@ def execute(self, userdata):
...
+            return "aborted"
+
+
+class SetFrameLookingAtState(smach.State):
+    def __init__(self, source_frame, look_at_frame, target_frame, duration_time=3.0):
+        smach.State.__init__(self, outcomes=["succeeded", "preempted", "aborted"])
+        self.source_frame = source_frame
+        self.look_at_frame = look_at_frame
+        self.target_frame = target_frame
+        self.duration_time = duration_time
```
eminmeydanoglu, line 470: a "duration" cannot fail to be "time". 
broadcast_duration ?

eminmeydanoglu (changes requested):
Good job. Close to perfect functionality.

I am leaving a review with some critical design and logic suggestions.

auv_smach/src/auv_smach/common.py
```diff
@@ -348,3 +608,84 @@ def execute(self, userdata) -> str:
...
+    def __init__(
+        self,
+        look_at_frame: str,
+        alignment_frame: str,
+        full_rotation: bool,
+        set_frame_duration: float,
+        source_frame: str = "taluy/base_link",
+        rotation_speed: float = 0.3,
+    ):
+        """
```
eminmeydanoglu, line 630: Reusable search state is cool

eminmeydanoglu (approved):
I do approve

PR #183, opened 2025-04-25
Add param use_sim_time = true for simulation 

eminmeydanoglu:


eminmeydanoglu:
This PR is tested and looks clean.

use_sim_time is already true without the added line, because there exists <arg name="use_sim_time" value="true" /> at start_pool_world.launch. Lets add it to main simulation launch too.

eminmeydanoglu:
@ozanhakantunca

ozanhakantunca (approved):

PR #186, opened 2025-04-28
Parameter to not start the autonomous system at simulation launch

eminmeydanoglu:
**New parameter:** `start_bridge` (default: `true`).  
If you plan to launch the autonomous system from a separate terminal, set `start_bridge: false` and run:

```bash
roslaunch auv_sim_bringup start_bridge.launch
```
No need to restart the simulation and RViz — and re-add all those messages (Hey @frk781)  — just because you fixed a typo!

Kudos to @frk781 for the idea

### DIFF
--- a/auv_sim/auv_sim_bringup/launch/inc/start_bridge.launch
+++ b/auv_sim/auv_sim_bringup/launch/inc/start_bridge.launch
@@ -2,7 +2,7 @@
 <launch>
   <arg name="namespace" default="taluy" />
   <arg name="control_rate" default="20" />
-  <arg name="use_gui" default="false" />
+  <arg name="use_gui" default="true" />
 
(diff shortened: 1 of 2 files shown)


ozanhakantunca (dismissed):
LGTM, we can merge this after a simulation test.

talhabw:
the errors caused because use_gui arg's default value was set to true in start_gazebo.launch but to false in start_bridge.launch. to avoid errors, they need to be set to same value when run separately.

I tried a few things with parameter server to get the value set in first launch file, but we can't turn the param to an arg in ros 1.

setting defaults to same value and keeping in mind that if we set it to false while running gazebo, you should set it to false in bridge too is good enough for us.

ozanhakantunca (approved):

PR #189, opened 2025-04-29
create initial torpedo task smach 

talhabw:
this is the initial smach for the torpedo task.

- `torpedo_frame_publisher.py` is responsible to create `torpedo_target` frame, which is the frame we launch torpedos from.
- smach simply follows this:
  - turn to torpedo map frame
  - follow a path to that frame
  - after the path is executed, wait a predefined amount of seconds while the align controller aligns to that frame
  - launch the torpedos


eminmeydanoglu:
Maybe as a new feature, we should make the align frame controller aware of whether the alignment is successfull. A simple threshold system (distance threshold + yaw threshold) would work. We could still keep the timeout but terminate the countdown if is_aligned.

talhabw:
i agree that `is_aligned` is something we should have. i'm not sure if it should be a feature of the Align Frame Controller though. instead, we could introduce a separate state that checks whether we're close enough to the target frame the Align Frame Controller is aligning to. this state could return `succeeded` when the alignment is good enough (within the limits it takes as a parameter, so we could have different preciseness requirements among different tasks). we can also handle the timeout logic within this state.

eminmeydanoglu:
What about a big "align frame" state with the sub-states;
1. Request alignment
2. Watch whether aligned (with a timeout)
3. Cancel aligment request (optional on off)

We could import this state whenever needed.

@talhabw

PR #190, opened 2025-04-29
fix follow path client

talhabw:
`execution_time` is not an existing value in the FollowPath action msg, I am guessing this was a thing in the early design and removed later on, and was forgotten here.

removing reference to that non-existing value was removed, so the issue was fixed

`FollowPath.action:`
```
# Goal
nav_msgs/Path[] paths

---
# Result
bool success

---
# Feedback
float32 current_segment_progress
float32 overall_progress
uint8 current_segment_index
```


eminmeydanoglu (approved):
Yep. Nice catch.

PR #191, opened 2025-04-30
Add Roll State

melihokur17:


ozanhakantunca:
Can you add the test reports here @melihokur17 ?

melihokur17:
When we tried it at the beginning of the pool test, it was successful, but over time, I think the sponges absorb water and the center of buoyancy shifts, so the robot cannot roll successfully. Over time, it becomes upright in the water and spins like a ballerina. In another branch, we tried a P D controller that would pub torque.y and torque.z to maintain the pitch and yaw angles of the vehicle at the beginning of the state, but tuning it is very difficult and the fact that it will need to be re-tuned for every mechanical change in the vehicle is a concern. Any suggestions @senceryazici ?

ozanhakantunca (approved):
Yeeyy

PR #196, opened 2025-05-17
Fix out of place log spams for object_map_tf_server_ros.cpp

eminmeydanoglu:


auv_navigation/auv_mapping/src/object_map_tf_server_ros.cpp
```diff
@@ -54,8 +54,8 @@ ObjectMapTFServerROS::ObjectMapTFServerROS(const ros::NodeHandle &nh)
       nh_.subscribe("object_transform_updates", 10,
                     &ObjectMapTFServerROS::dynamic_transform_callback, this);
 
-  ROS_INFO("ObjectMapTFServerROS initialized. Static frame: %s",
-           static_frame_.c_str());
+  ROS_DEBUG("ObjectMapTFServerROS initialized. Static frame: %s",
+            static_frame_.c_str());
 }
```
ozanhakantunca, line 59: We can keep this as ROS_INFO since it is called one time in a while and this is an information message :laughing:

auv_navigation/auv_mapping/src/object_map_tf_server_ros.cpp
```diff
@@ -111,8 +111,8 @@ bool ObjectMapTFServerROS::set_transform_handler(
         std::make_unique<ObjectPositionFilter>(*static_transform, 1.0 / rate_));
   }
 
-  ROS_INFO_STREAM("Stored static transform from " << static_frame_ << " to "
-                                                  << target_frame);
+  ROS_DEBUG_STREAM("Stored static transform from " << static_frame_ << " to "
+                                                   << target_frame);
```
ozanhakantunca, line 115: Same as the previous comment

ozanhakantunca (changes requested):
Hi @eminmeydanoglu I requested some changes. Maybe it's better to keep these as info message and adjust the neccessary ones as debug or debug_stream. Can you check?

auv_navigation/auv_mapping/src/object_map_tf_server_ros.cpp
talhabw: this is called for every transform created, which happens kind of a lot. I think we should keep this as debug to have cleaner logs in the bringup terminal

talhabw (approved):

ozanhakantunca (approved):

Issue #175, opened 2025-04-04
bug: poor timestamp assignment in odometry nodes (imu, dvl, pressure)


eminmeydanoglu:
All three odometry nodes are assigning the timestamp using rospy.Time.now() instead of using the timestamp from the original sensor message header. 

from imu_odometry_node.py:
```python
# Incorrect timestamp assignment
self.odom_msg.header.stamp = rospy.Time.now()

# Should be:
# self.odom_msg.header.stamp = imu_msg.header.stamp
```
Similar lines exist in dvl_odometry_node.py and pressure_odometry_node.py.

### Why this is problematic
This means we use the time the message is _created in this node_, not the time the actual IMU measurement was taken.
This introduces an avoidable timing inaccuracy. Sensor fusion nodes (our EKF) takes the timestamps of the messages (for precise timing) very seriously as far as I know, so correct timestamps are important. 

### On pressure data (/depth topic)
Both imu and dvl sensor data have stamped time, so we can directly pass them in. However the /depth topic is of type Float32, so no timestamps. My guess is this is just a poor message type choose, and we have the time data for the pressure sensor somewhere.

PR #202, opened 2025-05-24
 Use the timestamp from the IMU message instead of now

eminmeydanoglu:
About #175


eminmeydanoglu:
Lets try this in robot to see what is the time difference
 just for the science

eminmeydanoglu:
HARDWARE TEST SUCCEEDED!
ryan-gosling-thumbs-up

PR #205, opened 2025-05-26
Parallel Offset for Entrance and Exit frames

melihokur17:
To avoid hitting gates leg.

### DIFF
--- a/auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
+++ b/auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
@@ -38,7 +38,6 @@ class TransformServiceNode:
         self.exit_offset = rospy.get_param("~exit_offset", 1.0)
         self.z_offset = rospy.get_param("~z_offset", 0.5)
         self.parallel_shift_offset = rospy.get_param("~parallel_shift_offset", 0.15)
-        rospy.loginfo("entrance_offset = %f", self.entrance_offset)
         self.target_gate_frame = rospy.get_param(
             "~target_gate_frame", "gate_blue_arrow_link"
         )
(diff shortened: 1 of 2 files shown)


auv_navigation/auv_mapping/scripts/gate_trajectory_publisher.py
```diff
@@ -170,6 +221,29 @@ def create_trajectory_frames(self) -> None:
...
+                self.parallel_shift_offset,
+                self.tf_buffer,
+            )
+            exit_transform = self._shift_transform_parallel_to_gate_line(
+                exit_transform,
+                self.target_gate_frame,  # Frame in which exit_transform.transform is defined
+                other_gate_frame_for_shift_direction,
+                self.parallel_shift_offset,
+                self.tf_buffer,
+            )
```
eminmeydanoglu, line 245: These redundant lines could be avoided by making the helper function a "shift computer" instead of a "shifter". But not that big of a deal.

eminmeydanoglu (commented):
LGTM with just some small recommendations.
Lets try a pool test though.

talhabw:
btw, we need to remove the node name from the config file to make the params actually work, I think we can do it in this pr.

`gate_trajectory_publisher_params.yaml`:
```yaml
---
gate_frame_1: "gate_blue_arrow_link"
gate_frame_2: "gate_red_arrow_link"
target_gate_frame: "gate_blue_arrow_link"
entrance_offset: 1.0  # Distance (m) to offset the entrance frame along its normal direction
exit_offset: 1.0  # Distance (m) to offset the exit frame along the same normal direction
z_offset: 0.5  # Vertical drop below the gate frame (m)
parallel_shift_offset: 0.15  # Parallel shift offset (m)
```

eminmeydanoglu:
lets merge

ozanhakantunca (approved):
LGTM :DD

PR #207, opened 2025-05-31
create control inspector node

talhabw:
Other nodes/smach continues to publish /enable msg, but our main controller looks at `control_inspector_enable` topic to actually enable robot. 

`control_inspector_enable` is published by the new `control_inspector_node`. It ensures these safety checks passes before enabling the control:

- last odometry time
- last altitude time
- minimum altitude
- battery voltage
- last dvl time

we use depth to calculate altitude since dvl_altitude is not reliable for now.

this also means we can't move the robot without enabling robot, so something to consider before merging.


ozanhakantunca (approved):
LGTM at all :D We should wait for the next hardware test to merge this

PR #208, opened 2025-05-31
refactor launch files

talhabw:
move all launch and config files to auv_bringup.

also removed some legacy launch and config files that are not used anywhere in the codebase


ozanhakantunca (dismissed):
I couldn’t find any mistakes, but since this PR changes a lot of things, having multiple reviews is important to catch any potential oversights. And I think it needs a hardware test too. Nice work :D

PR #227, opened 2025-06-25
WIP : Dual Image Pipeline

ozanhakantunca:


ozanhakantunca:
Screenshot from 2025-06-25 23-52-35
Screenshot from 2025-06-25 23-51-20
These are the visualized yolo_results from the topics seperated from merged_image. I have written another node that takes the front/bottom image and draws bboxes from the front/bottom yolo_result message.

PR #235, opened 2025-06-27
Fixed full_rotation=false

melihokur17:


auv_smach/src/auv_smach/common.py
```diff
@@ -366,13 +366,12 @@ def normalize_angle(angle):
...
         try:
-            self.tf_buffer.lookup_transform(
+            return self.tf_buffer.can_transform(
                 self.source_frame,
                 self.look_at_frame,
                 rospy.Time(0),
-                rospy.Duration(1.0),
+                rospy.Duration(0.05),
             )
-            return True
```
eminmeydanoglu, line 375: can_transform makes sense. And we dont need the try except block anymore!

eminmeydanoglu (approved):

PR #240, opened 2025-06-30
Avoid checking inside image for  Botom cam

frk781:
Avoided checking for collisions with image edges to improve bin action decisions.


eminmeydanoglu (approved):
I trust this PR with my life

PR #241, opened 2025-06-30
feature: keep orientation for AFC

eminmeydanoglu:


auv_smach/src/auv_smach/common.py
```diff
@@ -179,11 +179,14 @@ def __init__(self):
 
 
 class SetAlignControllerTargetState(smach_ros.ServiceState):
-    def __init__(self, source_frame: str, target_frame: str):
+    def __init__(
+        self, source_frame: str, target_frame: str, keep_orientation: bool = False
+    ):
```
eminmeydanoglu, line 184: False by default

PR #254, opened 2025-07-03
fix odometry reset

### LINKED ISSUE #195: Bug: Unexpected behaviour from heading control
When control_enable is set to false, we feed the robot's current yaw state into cmd_pose_yaw. This should ensure that, when control_enable is later set to true, the robot maintains its current heading. However, we witnessed today, the robot instead attempts to rotate approximately 90 degrees to the left or right.

Let's investigate this.

talhabw:
closes #195 

Calling `set_pose` service to reset odometry with an default `PoseWithCovarianceStamped` results in `odometry yaw = 0`, but the `cmd_pose yaw` stays at the previous value, this causes weird rotations. 

We now have a `reset_odometry` service in `reference_pose_publisher`. Calling this service causes the heading control to act like robot is disabled for 2 secs after calling `set_pose`.

We can merge after hardware testing.


eminmeydanoglu (approved):
well-thought-out

auv_control/auv_control/launch/start.launch
```diff
@@ -45,6 +45,8 @@
       <remap from="cmd_vel" to="cmd_vel" />
       <remap from="cmd_pose" to="cmd_pose" />
       <remap from="enable" to="enable" />
+      <remap from="set_pose" to="set_pose" />
```
senceryazici, line 48: Is this set pose service supposed to be EKF node's set pose? If so how do we remap to that topic? Or is that topic namespace/set pose?

senceryazici (approved):
Code wise looks good to me, but I'm not sure how this fits into the system, so not commenting on that.

auv_control/auv_control/launch/start.launch
talhabw: Yes, the EKF node's set pose is `/namespace/set_pose`

PR #269, opened 2025-07-08
add D term to AlignFrame's P controller

talhabw:
we should test and tune it


auv_control/auv_control/scripts/align_frame_controller.py
```diff
@@ -22,18 +24,26 @@ def __init__(self) -> None:
         self.cmd_vel_pub = rospy.Publisher("cmd_vel", Twist, queue_size=10)
         self.enable_pub = rospy.Publisher("enable", Bool, queue_size=1)
 
-        self.rate = rospy.get_param("~rate", 0.1)
```
eminmeydanoglu, line 25: Haha, who would set it to 0.1 at first :D

PR #279, opened 2025-07-12
Restore Xsens IMU Integration

ozanhakantunca:


talhabw:
I created a seperate `imu_expansion_link`, so when we want to use bno as main imu, we can just change the remap in the `auv_bringup/launch/inc/boards/start_expansion_board_bridge.launch.xml` and disable the xsens without changing any links. We should probably change the `imu_expansion_link's` location before merging this.

frk781 (approved):
okay

PR #292, opened 2025-07-16
use path in torpedo smach

talhabw:


ozanhakantunca (dismissed):
Nice job! 😄

In out next step, I think we need to work on aligning torpedo_link with the torpedo_hole, possibly by adjusting our depth based on the z-position of torpedo_hole or torpedo_fire_link.
This could help improve our targeting accuracy before firing.

PR #293, opened 2025-07-17
New Coin Flip States

melihokur17:


ozanhakantunca (commented):
Nice job :D 
I think we could consider extracting the coin flip states into a separate SMACH file or adding a parameter to skip.
This would give us more flexibility. For example, in critical situations where we want to skip the coin flip logic entirely, we can easily initialize the SMACH sequence like:

init → gate → bin → torpedo

Or include it when needed:

init → coin → gate → bin → torpedo

This modular structure would also make our state machine cleaner and easier to maintain.
Let me know what you think :D

Shaun The Sheep

PR #301, opened 2025-07-18
make xsens default

ozanhakantunca:


talhabw:
we need to revert the [drift values](https://github.com/itu-auv/auv-software/commit/714925b3b9f74740ace390baf30e7b764f55a4c7#diff-ecd671e112120c437e8fedb52e940dfaf7ab8dbc35d01b80d873feeafe08df44) too

PR #304, opened 2025-07-19
Create a coefficient matrix for thrusters

ozanhakantunca:


senceryazici:
I remember that we already had some açalar for the direction of the thrusters no? Goes like


1.0, -1.0, 1.0, 1.0...

Can't this be also interpreted as the scalar?

2.0, -2.5, 3.0 etc?

PR #307, opened 2025-07-21
Add detection focus to gate

frk781:


ozanhakantunca (approved):
If it is tested we can merge :D

PR #362, opened 2025-07-30
Torpedo fire frames y offsets

melihokur17:


ozanhakantunca (approved):
LGTM, but we can fix the namings in torpedo.py also in this PR :D @melihokur17 
Shaun The Sheep

PR #372, opened 2025-08-06
[feature] yaw offset for a forward looking frame

eminmeydanoglu:
Introducing the frame forward_frame, which is computed by applying the offset "`wall_reference_yaw`" to odom orientation. Computations based on forward direction are now easier. No more need to dot product!


eminmeydanoglu:
Screenshot from 2025-08-06 05-12-58

PR #388, opened 2025-08-09
Use hypotenuse calculation for slalom pipes

ozanhakantunca:


eminmeydanoglu:
Thought about it before, likely won't work
But lets give it a try

PR #442, opened 2025-11-01
Visual slam (loop closure) with `rtabmap-slam`

eminmeydanoglu:
This is the WIP V-SLAM implementation of auv-software. :sunglasses:  For first steps, planning on having a loop closure system that will update map -> odom transform when a loop hypotesis is accepted. We will need to feed our local odometry. I was to test the loop closure system with Zedm + Zedm's visual odometry + rtabmap's visual odometry (yes it has one). Loop closure works reasonably, but of course lots of features in the world will be better as always. We will need to see the performance in underwater. 

rtabmap also works with RGB-M and even without external (or visual) odometry, but that will make the slam much less capable and slow. So don't feel like trying it with bags.

### DIFF
--- a/auv_navigation/auv_mapping/config/rtabmap_params.yaml
+++ b/auv_navigation/auv_mapping/config/rtabmap_params.yaml
@@ -14,8 +14,8 @@ subscribe_rgb: true
 subscribe_depth: true
 
 # External Odometry or VIO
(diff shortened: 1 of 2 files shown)


eminmeydanoglu:
Screenshot from 2025-11-01 15-10-19

PR #448, opened 2025-11-11
fix bags

talhabw:
- exclude `image_corrected`
- remove hardcoded topic name in the pattern

we should probably create the compressed version of `image_corrected` topic.

we also should follow the `x/image_raw` and `x/image_raw/compressed` naming conventions for image topics to avoid similar huge bag issues in the future


senceryazici:
> Can we check if this PR does what it is intended to do, and then we can merge quickly.
> 
> @talhabw 

Looking forward to it

PR #449, opened 2025-11-12
auto start localization service for simulations

### LINKED ISSUE #446: EKF localization node should start enabled in simulation
Currently, the EKF localization node is configured with disabled_at_startup: true in the launch file:

auv_bringup/launch/inc/localization/start_robot_localization.launch.xml
```<?xml version="1.0"?>
<launch>
  <arg name="namespace" default="taluy"/>
  <arg name="control_rate" default="20.0"/>
  <group ns="$(arg namespace)">
    <node pkg="robot_localization" type="ekf_localization_node" name="auv_localization_node" clear_params="true">
      <rosparam command="load" file="$(find auv_bringup)/config/ekf.yaml"/>
      <remap from="odometry/filtered" to="odometry"/>
      <remap from="accel/filtered" to="acceleration"/>
      <param name="disabled_at_startup" value="true" type="bool"/>
      <param name="frequency" value="$(arg control_rate)" type="double"/>
    </node>
  </group>
</launch>
```

This causes the node to be disabled at startup, requiring manual enabling. Let localization start automatically in any sim environment. 

So, set disabled_at_startup to false when running in simulation mode with a launch argument "sim", which shall be passed by start_gazebo.launch as true, and the default is false.

tixvage:
tries to fix #446 
Add `sim` as an argument to `start_gazebo.launch` and pass it on to simulation bridge and the other related components


eminmeydanoglu (dismissed):
Just works. Approved!

eminmeydanoglu:
Resolved some conflicts. Will be good to after the CI check. @tixvage

melihokur17 (approved):

PR #450, opened 2025-11-13
[WIP] Separate cmd_vel and cmd_pose timers from each others

tixvage:
Attempt to solve #115
I'm still not sure if this actually solves the issue, but from what I understand, when a disconnection happens, cmd_vel stays the same and never gets reset, so Taluy keeps moving. And this pr tries to solve it via separating timers. Still WIP and not tested in the pool yet, but it didn’t break anything in the simulation.

### DIFF
--- a/auv_control/auv_control/include/auv_control/controller_ros.h
+++ b/auv_control/auv_control/include/auv_control/controller_ros.h
@@ -136,11 +136,15 @@ class ControllerROS {
         continue;
       }
 
+      if (is_vel_timeouted()) {
+        desired_state_.tail(6).setZero();
+      }
+
       const auto control_output =
           controller_->control(state_, desired_state_, d_state_, dt);
 
       geometry_msgs::WrenchStamped wrench_msg;
-      if (is_control_enabled() && !is_timeouted()) {
+      if (is_control_enabled()) {
         wrench_msg.header.stamp = ros::Time::now();
         wrench_msg.header.frame_id = body_frame_;
         wrench_msg.wrench =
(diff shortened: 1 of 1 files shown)


eminmeydanoglu (approved):
Code looks good, and the sim test was successful. Chances are this will work in RL because why not. 
Though lets test it anyways. To test, we can use the wrench msg visualization again.

Lets wait for the hardware test to merge.

auv_control/auv_control/include/auv_control/controller_ros.h
```diff
@@ -136,11 +136,15 @@ class ControllerROS {
...
+      if (is_vel_timeouted()) {
+        desired_state_.tail(6).setZero();
+      }
+
       const auto control_output =
           controller_->control(state_, desired_state_, d_state_, dt);
 
       geometry_msgs::WrenchStamped wrench_msg;
-      if (is_control_enabled() && !is_timeouted()) {
+      if (is_control_enabled()) {
```
talhabw: i don't think setting desired velocity to 0 when `cmd_vel` timeouted is the correct way to do this.

this would result in controllers running actively to stop the robot. 
or if we were using the `cmd_pose` for x, y, z controls, a high enough output from the `cmd_pose pid` could cause the robot to just continue moving.

instead we can just stop publishing the `wrench`, just like how we did before with `!is_timeouted()`

also we don't even use the new `is_pose_timeouted`, which is not good.

maybe just replacing the old `!is_timeouted()` with `!is_pose_timeouted() && !is_vel_timeouted()` could be a solution?

talhabw: at the end of the day, to fix #115, we can just introduce separate timeouts for `cmd_vel` and `cmd_pose`, as you already did.

in another pr, i think we can develop to have a system to disable a single controller (`cmd_pose.x`, `cmd_vel.z`, etc.). 

that system could also help us with things like [heading control disable service](https://github.com/itu-auv/auv-software/blob/4fafedf8536248dac961c759d3aa731889800553/auv_control/auv_control/scripts/reference_pose_publisher.py#L114), which is implemented poorly.

a controller not getting new inputs, (aka. timing out) is an issue, not something we should use as a feature. which is what the #115 describes as a fix. 'if it timeouts, then robot should do x' is essentially introducing a side-effect, which could result in situations where we don't understand why something is happening.

PR #492, opened 2025-12-20
Pipe Follower Rework

tixvage:
still wip. Right now we can get different segments from camera view and get pixel locations of path. It doesn't have a proper launch structure for now because of lack of models etc...
Screenshot From 2025-12-20 17-04-53


tixvage:
right now, the code fully works in simulation. it takes nearly less then 4 minutes to finish and go back from same path to where it started. still needs some tweaks, also integration to our smach system and launch files

tixvage:
roslaunch auv_sim_bringup start_gazebo.launch world:=tac_sea

roslaunch auv_detection tracker_segment.launch device:="cuda:cpu"

roslaunch auv_smach start.launch test_mode:=1 test_states:=init,pipe

auv_smach/src/auv_smach/pipe_follower.py
```diff
@@ -0,0 +1,95 @@
...
+                    confirm_duration=5.0,
+                    timeout=60.0,
+                    cancel_on_success=True,
+                ),
+                transitions={
+                    "succeeded": "DISABLE_PUBLISHER",
+                    "preempted": "DISABLE_PUBLISHER",
+                    "aborted": "DISABLE_PUBLISHER",
+                },
+            )
```
eminmeydanoglu: potential issue: cancel_on_success=true. what if taluy catches the carrot? the distance between them is <0.2 for 5 seconds: align frame is cancelled and smach continues. but the pipe is not completed yet!

question: is it possible for the robot and carrot to get quite close to each other?

for example look at DynamicPathState: there's no threshold check but it's not a problem, because the carrot frame there is guaranteed to stay 1 meter ahead of the robot. the robot and carrot can only get close when reaching the last point of the path.


AlignFrame is actually designed for static targets. maybe somehow pipe_frame_publisher should send a pipe completed signal? or if not aligning is guarenteed like in dynamicpath, we are good.

eminmeydanoglu (changes requested):
Solid! Leaving some comments but only the long ones are important.

frk781:
baya kötü olmuş :-1:

PR #610, opened 2026-03-31
slalom object map tf server

tixvage:
slalom with object map tf server and camera pos estimator instead of custom clustering methods
```cmd
roslaunch auv_smach start.launch test_mode:=1 test_states:=init,gate,slalom_exp namespace:=taluy roll:=0 yaw:=0 pitch:=0
```
 takes nearly 3 minutes


tixvage:
#661

PR #627, opened 2026-04-10
Fix stale seg timestamps in bottom_seg pipeline

eminmeydanoglu:


auv_vision/auv_vision/scripts/handlers/segment_camera_handler.py
```diff
@@ -39,15 +39,8 @@ def __init__(
...
 
-    def handle(self, detection_msg: YoloResult):
-        seg = self.shared_state.get("last_segment_measurement")
-
-        # Use segment measurement timestamp if available, otherwise detection stamp
-        # ı did this beacuse ı do not want to abandon the bottle frame completly.
-        if seg is not None and seg.valid:
-            stamp = seg.header.stamp
-        else:
-            stamp = detection_msg.header.stamp
```
eminmeydanoglu, line 50: The detection timestamp is the right source of truth. No point using the timestamp after masking.

PR #632, opened 2026-04-12
Improve/fix TF timestamps across control, mapping, smach

eminmeydanoglu:
These are a set of improvements/ bug fixes, and all of them are about timestamps. #627 and #628 hopefully adress the errors and warnings we saw. This PR builds on them to make the system more robust and more semantically correct. I am commenting on the code changes below, follow there

Did a full smach run, everything seems fine

### DIFF
--- a/auv_control/auv_control/include/auv_control/controller_ros.h
+++ b/auv_control/auv_control/include/auv_control/controller_ros.h
@@ -228,6 +228,7 @@ class ControllerROS {
   void cmd_pose_callback(const geometry_msgs::PoseStamped::ConstPtr& msg) {
     const auto source_frame = get_source_frame(msg->header.frame_id);
     auto transformed_pose = msg->pose;
+    const auto lookup_time = msg->header.stamp;
 
     static tf2_ros::Buffer tf_buffer;
     static tf2_ros::TransformListener tf_listener(tf_buffer);
@@ -239,7 +240,7 @@ class ControllerROS {
                 depth_control_reference_frame_.c_str());
       try {
         transform_stamped = tf_buffer.lookupTransform(
-            depth_control_reference_frame_, source_frame.value(), ros::Time(0),
+            depth_control_reference_frame_, source_frame.value(), lookup_time,
             ros::Duration(transform_timeout_));
 
         tf2::doTransform(msg->pose, transformed_pose, transform_stamped);
@@ -256,7 +257,8 @@ class ControllerROS {
     desired_state_.head(6) = auv::common::conversions::convert<
         geometry_msgs::Pose, ControllerBase::Vector>(transformed_pose);
 
-    latest_cmd_pose_time_ = ros::Time::now();
+    latest_cmd_pose_time_ =
+        lookup_time.isZero() ? ros::Time::now() : lookup_time;
   }
 
   void accel_callback(
--- a/auv_control/auv_control/scripts/reference_pose_publisher.py
+++ b/auv_control/auv_control/scripts/reference_pose_publisher.py
@@ -32,6 +32,7 @@ import numpy as np
 
 import dynamic_reconfigure.client
 from auv_common_lib.control.enable_state import ControlEnableHandler
+from auv_common_lib.transform import is_transform_fresh
 from threading import Lock
 from tf2_geometry_msgs import do_transform_point, do_transform_vector3
 from geometry_msgs.msg import Vector3Stamped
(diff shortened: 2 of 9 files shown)


auv_control/auv_control/include/auv_control/controller_ros.h
```diff
@@ -239,7 +240,7 @@ class ControllerROS {
                 depth_control_reference_frame_.c_str());
       try {
         transform_stamped = tf_buffer.lookupTransform(
-            depth_control_reference_frame_, source_frame.value(), ros::Time(0),
+            depth_control_reference_frame_, source_frame.value(), lookup_time,
             ros::Duration(transform_timeout_));
 
         tf2::doTransform(msg->pose, transformed_pose, transform_stamped);
```
eminmeydanoglu, line 246: yet another sync issue: cmd_pose was transformed with time(0), so controller could use a newer tf than the one cmd_pose was computed with. generally the target frames are very similar to each other (because they dont have velocity in the world), nonetheless they move in the odom frame since kalman filter updates them.

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -663,14 +696,35 @@ def control_loop(self):
...
+            )
+            if target_to_odom is None:
+                rospy.logwarn_throttle(
+                    1.0, f"Failed to lookup target frame '{frame_id}'"
+                )
+                return
+            cmd_pose_stamp = target_to_odom.header.stamp
+
+        cmd_pose_stamped.header.stamp = cmd_pose_stamp
+
```
eminmeydanoglu, line 717: if target frame != odom, use the latest target->odom stamp as the time anchor for cmd_pose. now the depth/orientation handlers and the controller all use the same tf snapshot. before this, cmd_pose was stamped as now, while parts of it were computed with ros.time(0), so we mixed different tf samples.

auv_navigation/auv_mapping/src/object_map_tf_server_ros.cpp
```diff
@@ -340,10 +340,11 @@ ObjectMapTFServerROS::transform_to_static_frame(
 std::optional<geometry_msgs::TransformStamped>
 ObjectMapTFServerROS::get_transform(const std::string &target_frame,
                                     const std::string &source_frame,
+                                    const ros::Time &lookup_time,
                                     const ros::Duration timeout) {
   try {
     auto transform = tf_buffer_.lookupTransform(target_frame, source_frame,
-                                                ros::Time(0), timeout);
+                                                lookup_time, timeout);
     return transform;
```
eminmeydanoglu, line 348: Syncing issue again

auv_control/auv_control/scripts/reference_pose_publisher.py
```diff
@@ -622,13 +643,25 @@ def tf_lookup(
...
+            if time == rospy.Time(0) and not is_transform_fresh(
+                transform, self.tf_freshness_threshold
+            ):
+                age = abs((rospy.Time.now() - transform.header.stamp).to_sec())
+                rospy.logwarn_throttle(
+                    1.0,
+                    f"TF lookup from {source_frame} to {target_frame} returned stale data "
+                    f"({age:.3f}s > {self.tf_freshness_threshold.to_sec():.3f}s)",
+                )
+                return None
```
eminmeydanoglu, line 655: Freshness check! Rospy.Time(0) is just the last available stamp. But this is a controller and it is meant to use real time stamps to compute errors. Why not use rospy.time.now? Find and ask me :D

auv_smach/src/auv_smach/common.py
```diff
@@ -991,8 +994,12 @@ def __init__(
...
     def get_error(self):
         try:
-            transform = self.tf_buffer.lookup_transform(
-                self.source_frame, self.target_frame, rospy.Time(0), rospy.Duration(4.0)
+            transform = lookup_fresh_transform(
+                self.tf_buffer,
+                self.source_frame,
+                self.target_frame,
+                rospy.Duration(rospy.get_param("~tf_lookup_timeout", 0.2)),
+                rospy.Duration(rospy.get_param("~tf_freshness_threshold", 0.2)),
```
eminmeydanoglu, line 1002: Again, RotationState is meant to check if a transform is available RIGHT NOW, and time(0) is not always that. Freshness check.

talhabw (approved):

PR #639, opened 2026-04-16
fix min distance logic in object map tf server

tixvage:


frk781 (approved):
looks solid, fixes the old threshold logic

PR #666, opened 2026-05-09
Add torpedo check inside image

frk781:


tixvage:
implemented with #659

PR #670, opened 2026-05-16
Accelerate gazebo again

eminmeydanoglu:
## What changed?

- Added the `pointcloud_filter` argument. When set to `0`, the simulated Realsense D435 is not added to the Taluy URDF.
- Added the `camera_update_rate` argument. Logitech camera sensors now default to `10 Hz` instead of fixed `30 Hz`.
- Changed `pool.world` physics from `max_step_size=0.002`, `real_time_update_rate=500` to `0.01`, `100`.
- Changed the pool overhead camera from `30 Hz` to `10 Hz`.
- Decoupled RViz from `use_gui`; `use_rviz` now defaults to `false`.
- Disabled `tf_gazebo_debug` by default; enable it with `start_tf_gazebo_debug:=1`.
- Set `robot_state_publisher` to `respawn="false"` and removed its unnecessary private `robot_description` parameter.

## Why changed?

- Full-stack startup RTF had dropped to about `0.43-0.45`.
- Pointcloud, camera consumers, small physics step size, RViz, and debug nodes added unnecessary default load.
- `robot_state_publisher` sometimes entered a respawn loop and added process churn.

## Metric impact

| Case | Result |
| --- | --- |
| Full change set | RTF improved from about `0.43-0.45` to the `~1.0` range. |
| Bridge on, bbox on, camera `10 Hz`, pointcloud off | Median RTF `~1.000`. |
| Bridge on, bbox on, camera `30 Hz`, pointcloud off | Median RTF `~0.856`. |
| Gazebo-only, camera `10 Hz`, pointcloud off | Median RTF `1.017`, `gzserver` about `37%` CPU. |
| Gazebo-only, camera `30 Hz`, pointcloud off | Median RTF `1.014`, `gzserver` about `38%` CPU. |
| Gazebo-only, pointcloud on | `gzserver` about `47-51%` CPU. |
| Pointcloud topic bandwidth | `/points_filtered` about `129-299 MB/s`. |


eminmeydanoglu:
@tixvage willl work on objects and object physics.

